PR #1, opened 2021-07-28
Add listen paths config

kirillplatonov:
First of all, thank you for the amazing gem. 👏 I started using it for https://github.com/baoagency/polaris_view_components and it works great so far! It really makes the process of components development much easier.

In this PR I added additional configuration option to allow auto-reload on component changes:
```ruby
config.lookbook.listen_paths << Rails.root.join('../app/components')
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-<div align="center"> 
+<div align="center">
 <h2> 👀 Lookbook 👀</h2>
- 
+
 ⚡️ Supercharge your [ViewComponent](http://viewcomponent.org/) development process 🚀
 
 </div>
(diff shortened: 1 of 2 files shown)


allmarkedup:
Hey @kirillplatonov thanks for this, looks great. I think it would actually make sense to include the view_component components path (`config.view_component.view_component_path`) by default anyway to be honest. And then users can use this config option to additionally add any other paths on top of that. 

I'll give this PR a test when I have a chance later and then merge it in. Then I'll make the change to add the components path as a default once that is done.

Very much appreciate the contribution, thanks again!

kirillplatonov:
Sounds good!

allmarkedup:
Work great. Merged 👍

PR #12, opened 2021-07-30
Respect selected preview controller

kirillplatonov:
We use a custom view component controller to load custom view helpers in previews.


afomera:
Just ran into this, and having this functionality would be really great, as we have hundreds of components that use app helpers 😭

allmarkedup:
@kirillplatonov ah yes of course, should definitely use the custom preview controller if one has been specified. Good catch, thanks!

PR #14, opened 2021-08-03
Fix rake tasks

kirillplatonov:
Rake tasks in the main app are broken right now:
CleanShot 2021-08-03 at 22 57 49@2x


allmarkedup:
@kirillplatonov thanks for this!

However the `require` statement is still needed in order for the tasks to be runnable from within the Lookbook codebase (as opposed to only from the parent app). Changing it to `require_relative "../lookbook.rb"` rather than removing it should make it runnable from both places I think?

allmarkedup:
@kirillplatonov perfect, thank you 👍

PR #35, opened 2021-08-25
Enable compatibility with Rails 5.x

edwinthinks:
Hey there @allmarkedup!

I noticed that this gem only worked for Rails 6.x but one of my projects is still at Rails 5.x. This PR is my attempt at making this library usable on projects that are still on Rails 5.x. I've tested this on my project Rails 5.x and was able to get it to work!

I think it is reasonable to support Rails 5.x since the view component library does too?

Here are few caveats that I haven't figured out yet.
- I'am not sure how to get auto-refresh working. I do notice that the refresh events are firing in the terminal but the browser doesn't reload. (Note -- I usually use some webpacker configurations)
- (Maybe unrelated) Even if I turn off auto_refresh, I do see events like this in the terminal:
```
10:00:42 web.1             | Started GET "/lookbook/cable?uid=1629903490194" for ::1 at 2021-08-25 10:00:42 -0500
10:00:42 web.1             | Processing by Lookbook::BrowserController#show as HTML
10:00:42 web.1             |   Parameters: {"uid"=>"1629903490194", "path"=>"cable"}
10:00:42 web.1             |   Rendering /Users/emak/projects/lookbook/app/views/lookbook/browser/not_found.html.erb within layouts/app
10:00:42 web.1             |   Rendered /Users/emak/projects/lookbook/app/views/lookbook/browser/not_found.html.erb within layouts/app (0.7ms)
10:00:42 web.1             |   Rendered /Users/emak/projects/lookbook/app/views/lookbook/partials/nav/_label.html.erb (0.3ms)
10:00:42 web.1             |   Rendered /Users/emak/projects/lookbook/app/views/lookbook/partials/nav/_preview.html.erb (2.4ms)
10:00:42 web.1             |   Rendered /Users/emak/projects/lookbook/app/views/lookbook/partials/nav/_nav.html.erb (4.4ms)
10:00:42 web.1             |   Rendered /Users/emak/projects/lookbook/app/views/lookbook/partials/_sidebar.html.erb (6.2ms)
10:00:42 web.1             |   Rendered /Users/emak/projects/lookbook/app/views/lookbook/partials/_icon_sprite.html.erb (1.4ms)
10:00:42 web.1             | Completed 200 OK in 16ms (Views: 11.8ms)
```

Please let me know what you think. I'd love to add some tests to this :)


allmarkedup:
Hi @edwinthinks - sorry it's taken me so long to look into this further.

I've done some testing and everything apart from the auto-refreshing of the UI seems to work with Rails 5.x. The issue with the auto refresh seems to be on the client side - Rails 6.x uses the `@rails/actioncable` npm module whereas Rails 5.x used some JS that was bundled with ActionCable and included via Sprockets. Lookbook is using the `@rails/actioncable` package to handle the client-side connection and that doesn't seem to be compatible with the way Rails 5.x ActionCable messages are sent.

There may be a way to patch this (or to include a Rails 5.x ActionCable-compatible third party library like https://github.com/mwalsher/actioncable-js if dependent on a version check) but it will require a little more work.

I may merge your PR to relax the Rails version constraint to support v5.x apps anyway and just add a note about the auto-refresh in the docs, as it's not really a show-stopper if people are keen to use Lookbook and are happy to live without the auto-refresh for now. Would you be happy with that?

edwinthinks:
@allmarkedup I think that sounds great!

allmarkedup:
@edwinthinks I've just released v0.3.1 that enables Rails v5.x compatability, and added a note in the docs about the auto-refresh not being supported in versions less than 6.

Give it a spin and let me know if you run into any issues. And thanks again for the PR! :-)

PR #40, opened 2021-09-03
Delete .DS_Store

Spone:
I think this file should not be committed :)


allmarkedup:
Ha you are right! Good spot, thanks :-)

PR #41, opened 2021-09-05
Relax the gemspec to work with Rails 7

afomera:
Without this, Lookbook fails to allow the app to bundle when running Rails from github and soon to be released the 7.0 alpha gems.

We dual-boot our rails application at work with two versions of Rails (Latest released 6.1 and then main from GitHub) and ever since the dependency was locked to ~> 6.0 this has prevented us from bundling.

I don't think there's anything specific in Lookbook that would break from Rails 7 but we'll certainly keep our eye out.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     lookbook (0.2.1)
       listen (~> 3.3)
-      rails (~> 6.0)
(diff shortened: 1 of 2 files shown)


allmarkedup:
Sure, sounds good! If you run into any issues just let me know :-)

PR #53, opened 2021-11-17
Enable previews to work with ViewComponentContrib

liaden:
ViewComponentContrib allows for component_name/preview.rb file structure. Changing the glob allows for lookbook to work, as long as the preview does not inherit from ViewComponentContrib::Preview (due to conflicting Preview templates and who does the rendering).


allmarkedup:
Thanks @liaden - I've been meaning to look into ViewComponentContrib compatibility so this is a great start.

When I have a little time I'll try and see if there is any way to get Lookbook working even when using `ViewComponentContrib::Preview` but until I've dug in a bit more I'm not sure what that would involve or if it's even possible.

Thanks again 👍

PR #54, opened 2021-12-13
Stop rendering `srcdoc` preview when it is disabled

czj:
As [mentioned in documentation](https://github.com/allmarkedup/lookbook#blank-preview-window) and handled in configuration, setting `config.preview_srcdoc` should disable `srcdoc` content generation.

The controller did not take into account this flag and always rendered it and made it accessible by the view.

This PR fixes this behavior.

I'm willing to also check in the view if `@preview_srcdoc` is blank to skip outputting `srcdoc=""` altogether.

### DIFF
--- a/app/controllers/lookbook/app_controller.rb
+++ b/app/controllers/lookbook/app_controller.rb
@@ -27,7 +27,9 @@ module Lookbook
         begin
           set_params
           @examples = examples_data
-          @preview_srcdoc = render_examples(examples_data).gsub("\"", "&quot;")
+          @preview_srcdoc = if lookbook.preview_srcdoc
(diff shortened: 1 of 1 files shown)


allmarkedup:
Hi @czj and thanks for this!

This is a regression that crept into the last release I put out I believe, good catch. Previously there was a check for the setting in the view which removed the srcdoc attribute when appropriate. I think that should be re-instated (in addition to the controller check you have put in here), rather than just outputting an empty srcdoc value.

I need to make some other updates to the views so I'll merge this in and then add the view check (and a test for this behaviour) with the other changes. I have a busy few days unfortunately so a release will probably need to wait until the end of the week.

Many thanks for your help :-)

PR #58, opened 2021-12-24
Update instructions for mounting in production mode

robzolkos:
The README is not too clear on how to configure things to view lookbook in production (there are two settings that need to be changed from defaults to make it work).

This update makes those two settings/changes clearer (it had me stumped for a while till I figured it out).


allmarkedup:
Thanks for this @robzolkos! Much appreciated and sorry you got stuck on it.

PR #60, opened 2022-01-20
Better visualization for preview named Preview 

muriloime:
Hi, first of all congratutations. This is a truly gem  :nerd_face: 

I tend to prefer the convention as per https://github.com/palkan/view_component-contrib where components are named Component , previews are named Preview and we put everything on the same folder. This small commit helps with the visualization of such convention, otherwise I get a list of folders named preview on my nav. 

Thanks and regards 

Murilo

### DIFF
--- a/lib/lookbook/preview.rb
+++ b/lib/lookbook/preview.rb
@@ -44,7 +44,7 @@ module Lookbook
     # Examples::FooBarComponentPreview -> "Examples::FooBar"
     # Examples::FooBarComponent::Preview -> "Examples::FooBar"
     def lookbook_name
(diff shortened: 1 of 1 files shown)


allmarkedup:
Hey @muriloime - many thanks for this and very sorry about the slow reply, I've had a busy few weeks!

This looks great, I'll merge it in now. I'm hoping to do a release later today once I fix an outstanding bug so I'll include it in that.

Thanks again for the contribution :-)

allmarkedup:
@muriloime I've just release v0.5.2 with this change included in it. Appreciate the contribution 👍

PR #62, opened 2022-02-16
Replace `rails` dependency with `actioncable` + `railties`

nickcampbell18:
By making `rails` a dependency, it means you can't use this project in applications using only parts of Rails (e.g. API-only apps, or those which don't use sprockets, mail, etc). As a gem we should try and require only the bare minimum required to function. [Here's another example where requiring the whole of Rails was a problem](https://twitter.com/nateberkopec/status/1374722404228853762).

I believe we only need actioncable and railties for this engine to function properly. It also looked like activestorage isn't actually used anywhere so I've also removed that, but I can easily put that back as a development-only dependency if preferred.

I'm using this version of lookbook locally and it hasn't caused any problems following this change.


allmarkedup:
This is great, thanks @nickcampbell18 - I'll try to give it a test as soon as I can to make sure there are no unintended consequences but once if all is good I'll merge it in. Thanks again for the contribution!

nickcampbell18:
Thank you kindly! Let me know if you need any further help.

allmarkedup:
Hey @nickcampbell18, I've just given this a spin and it all seems to be working great so I'm going to merge this in now and will include the changes in the next release. Much appreciated 👍

PR #69, opened 2022-03-30
Alllow class name to be  simply Preview

muriloime:
This commit allows Preview class names to be like both  SamplePreview and Sample::Preview


allmarkedup:
Looks good, thanks @muriloime :-)

PR #70, opened 2022-03-30
Fixes exception when raising error 😜

muriloime:
This fixes exception when raising errors with ruby version > 3 . < 3 untested

### DIFF
--- a/app/controllers/lookbook/previews_controller.rb
+++ b/app/controllers/lookbook/previews_controller.rb
@@ -176,7 +176,7 @@ module Lookbook
           }
         end
       end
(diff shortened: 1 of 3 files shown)


stevegeek:
This sort of fix for Ruby 3 is necessary also at https://github.com/allmarkedup/lookbook/blob/main/app/helpers/lookbook/component_helper.rb#L13 (splat the hash to keyword args)

allmarkedup:
Thanks guys - I need to get some better testing set up to catch things like this with different Ruby versions! Much appreciated (and will fix the component helper in a separate commit).

PR #78, opened 2022-04-08
Fix not scrollable overflowing long params list

mdrbohlav:
Fix for when the list of parameters in the drawer is too long and overflowing and the panel is not scrollable.


allmarkedup:
Thanks @mdrbohlav! Good catch. Merging now 👍

PR #80, opened 2022-04-11
Update README.md

dmorgan-fa:
Wrap the colour example in quotes, otherwise the string is interpreted as a comment and rails fails to start.


allmarkedup:
Good catch! Thanks @dmorgan-fa 👍

PR #97, opened 2022-06-08
Adding tabs to pages

leighhalliday:
Currently our team at Wrapbook uses Lookbook for our ViewComponent library. We have a separate "Style Guide" and also keep usage details/documentation for designers in Figma. Our goal is to move all of this into Lookbook Pages. Because there are different audiences using the Design System:

- Designer
- Frontend Engineer
- Other Engineers
- Mobile Engineers

We wanted to split up content into separate tabs, using Shopify's Polaris as an example, as can be seen on their [Avatar](https://polaris.shopify.com/components/images-and-icons/avatar) page.

This PR provides an opt-in way to add Tabs to Pages, and it is done within separate files (with similar names) that are detected and then added within a Page.

Example with code:
image (6)

Example with embedded preview:
image (7)

File naming uses square brackets to indicate you want it as a tab on an existing page. If the underlying page does not exist, the tab is ignored.

```
test/components/docs/
  ├── 01_avatar.md.erb
  ├── 01_avatar[design].md.erb
  ├── 01_avatar[mobile].md.erb
  ├── 01_avatar[web].md.erb
```

Would love your feedback!


lib/lookbook/utils.rb
```diff
@@ -6,7 +6,7 @@ module Utils
     protected
 
     def generate_id(*args)
-      parts = args.map { |arg| arg.to_s.parameterize.underscore }
+      parts = args.map { |arg| arg.to_s.force_encoding("UTF-8").parameterize.underscore }
```
leighhalliday, line 9: Was getting `ASCII-8BIT` errors when embedding previews in the dummy app, not just on Tabs but on any Page... perhaps this was my own issue, but adding this fixed the issue.

app/views/lookbook/components/_embed.html.erb
```diff
@@ -1,4 +1,4 @@
-<div id="<%= id %>" x-data="embed" class="not-prose embed border border-gray-300 rounded-md overflow-hidden" @page:morphed.window="recaclulateIframeHeight">
+<div id="<%= id %>" x-data="embed" class="not-prose embed border border-gray-300 rounded-md overflow-hidden" @page:morphed.window="recaclulateIframeHeight" @tab-switch.window="recaclulateIframeHeight">
```
leighhalliday, line 1: Embedded previews on hidden tabs need their height recalculated when visible for the first time (otherwise they hide most of the content)... this event is emitted when tabs are changed.

allmarkedup:
Hey @leighhalliday @sakin thank you so much for this! I've been off for a few days so I'm only just getting round to having a look at it. I'm going to dig into it a bit more today/tomorrow but in principle I think it looks great. 

Really appreciate your work on this and I'll let you know once I've been able to have a look in more detail.

allmarkedup:
@leighhalliday okay so regarding this PR, I've taken a look and generally really like it - especially how you've made sure each tab is still it's own `Page` so supports all the YAML metadata etc. I have spotted a couple of minor issues which should be easily fixed:

1) [This line](https://github.com/leighhalliday/lookbook/blob/79ff8b03dc2f330ed4bddb6815664c92e9920e42/lib/lookbook/page.rb#L131) could be changed to:
```ruby
@options[:label] ||= (remove_position_prefix(tab? ? tab : path_name)).titleize
```
To ensure that tabs have the expected label even if not set manually in the YAML frontmatter

and 2) The tabs do not currently support the (I think undocumented?!) `position` option in the YAML which allow pages to be given an index position without having to use the file name prefix. This is probably more useful for tabs though so that they don't always need to be in alphabetical order. It can be easily fixed by changing [these lines](https://github.com/leighhalliday/lookbook/blob/79ff8b03dc2f330ed4bddb6815664c92e9920e42/lib/lookbook/page.rb#L171-L173) to something like:
```ruby
page_dict = sorted_pages.index_by(&:path)
sorted_tabs = tabs.sort_by { |tab| [tab.position, tab.label] }

sorted_tabs.each do |tab|
# etc...
```

Are you happy to make those changes to your PR? Then I'd be happy to merge this.

---

As for future contributions - I definitely think some sort of draft PR or even just an issue outlining the feature you want to add ahead of time would be good, just so I can give you a heads up if anything like that is already planned or in the works.

I'm actually trying to focus most of my time at the moment on getting a [v1.0 release](https://github.com/allmarkedup/lookbook/issues/96) ready which will contain quite a few internal changes, so it makes even more sense at this point I think to have a brainstorm before you write too much code.

For example, I'm shortly going to be including an update to the v1.0 branch that will allow some customisation/addition of the preview page tabs which will most likely make building the accessibility integration you mention (which I love the idea of and has been requested before!) much easier. And this [feature request](https://github.com/allmarkedup/lookbook/issues/75) will also be shipped as part of the v1.0 release which may actually include something like the `embed_code` helper you suggest too.

I will port over this tabs functionality (possibly with a few additions) to the v1.0-beta once this is merged into the current branch.

Thank you so much again for all your work on this and I look forward to collaborating on new features in the future :-)

allmarkedup:
Fantastic, thanks @leighhalliday :-)

leighhalliday:
@allmarkedup I've made the changes (thank you for the suggestions!!) and added tests for them, they're in this commit: https://github.com/allmarkedup/lookbook/pull/97/commits/c97a3cd4eed77ff49b9cc55343648dc6f336d0ed

leighhalliday:
ps - Sorry I did this against main branch... I realize I should have done this against the beta branch! Please let me know if you want any help getting this into the beta branch.

allmarkedup:
@leighhalliday no, main branch is fine - the beta is realistically still a while away from a proper release I think so would be good to get it out and then I can port it over to the v1.0 branch manually as I may make a few changes/additions.

For example, I think if there is the concept of tabs in pages now it would be nice to have a helper to let people create tabs inline within the page, without needing external files (and then the external-file tabs can basically use that helper under the hood). I don't think that is needed for a first release but might be nice addition to the 1.0 beta.

I'm going to do a last check of this now and then hopefully merge it in and get a v0.9.0 release out tonight. Very cool, thank you :-)

allmarkedup:
@leighhalliday this is all working great so I've just released it as part of v0.9.0 🎉 😊

allmarkedup:
@leighhalliday and just a heads up I've now ported over the tabs functionality to the v1.0-beta branch and released a new beta with it in - **v1.0.beta.1**.

If you get a chance to play with it then just let me know if you find any issues. The implementation is a bit different but the way it works should be identical. Thanks again :-)

allmarkedup:
@leighhalliday ok that is great - be aware it probably still has a lot of rough edges though! I've got a lot of tests to catch up on :-)

I've just released a new beta version (`beta.3`) and updated the docs as this now provides the option of adding your own custom inspector/preview tabs. I'd love to hear your feedback if you have a chance to have a play and see if it might help you to add in some of the features that you mentioned before.

And any problems getting up and running on the beta just give me a shout and I'll try to help debug :-)

PR #108, opened 2022-06-30
fix: Escape quotes in params values

baygeldin:
Hi! Thanks for the awesome gem. There was a minor bug that prevented using params with quotes, e.g.:

```ruby
  # @param message text
  def example(message: "Let's go!")
    # render component
  end
```

image


allmarkedup:
Ah great catch - thanks @baygeldin! Really appreciate the PR. All looks good to me so I'll get this merged in now and get a release out with the fix in it shortly. Thanks again :-)

PR #114, opened 2022-07-18
Replaced max-w-3xl with max-w-screen-lg

stevetidy:
I found that the width for docs content was too narrow. I added a different class name to make it wider:

Screen Shot 2022-07-18 at 2 12 42 PM

### DIFF
--- a/app/views/lookbook/pages/show.html.erb
+++ b/app/views/lookbook/pages/show.html.erb
@@ -1,5 +1,5 @@
 <div class="px-4 md:px-10 pt-8 md:pt-10 overflow-auto scroll-smooth w-full max-h-full pb-12" x-ref="scroller">
-  <div class="w-full max-w-3xl mx-auto h-full flex flex-col">
+  <div class="w-full max-w-screen-lg mx-auto h-full flex flex-col">
     <% if @page.header? %>
       <header id="page-header" class="mb-8 prose max-w-none flex-none">
         <h1><%= @page.title %></h1>


allmarkedup:
Thanks for this @stevetidy. I have a bit of a reservation about this as I think it may make line lengths for text content too long for comfortable reading.

It feels like this should maybe be a config setting - or perhaps that elements such as embedded previews (such as in your screenshot) should be allowed to stretch wider than the general prose width.

I'll leave this open for now as I'd like to have a think about it before just pushing the width of the whole area wider. Let me know what you think too if you have time :-)

stevetidy:
> Thanks for this @stevetidy. I have a bit of a reservation about this as I think it may make line lengths for text content too long for comfortable reading.
> 
> It feels like this should maybe be a config setting - or perhaps that elements such as embedded previews (such as in your screenshot) should be allowed to stretch wider than the general prose width.
> 
> I'll leave this open for now as I'd like to have a think about it before just pushing the width of the whole area wider. Let me know what you think too if you have time :-)

Hi @allmarkedup, I didn't personally find any issue with the width of the lines adversely affecting one's ability to read, but I assume that's something that can affect people differently. There's certainly sites with  fluid widths that have no limitation on the width of the reading area. Wikipedia is the most prominent example that comes to mind.

The width of the embed was certainly the biggest issue for our team. There's a few components and code example that render the the `mobile` view rather than the `desktop` view because the embed window is so narrow. It gives the impression to designers that there may be a bug that needs to be fixed.

I guess either allowing the embed code to be much wider or allowing for the ability to configure/customize the width(s) of the docs pages would be the way to go.

Thanks for responding though 😄

allmarkedup:
@stevetidy yeah I also personally don't have too much of a problem with longer lines but I know that a lot of folks prefer to try and keep line lengths between 45 to 80 characters (partly also for accessibility reasons, I believe). So would be good to come up with a middle-ground solution I think.

Totally agree with you on the need for the embed (and other non-prose elements) to be wider though - not being able to preview the 'desktop' views of components embedded in the docs is definitely not ideal!

I'm going to have a look at this now and see if I can come up with a solution that works for all cases.

allmarkedup:
@stevetidy I've changed my mind :-) I've had a look at your change and although the line lengths are a little long I think they are still pretty readable. The embeds definitely need be wider though so I'm going to merge in your PR and try to look at constraining line lengths when I have a little more time.

Thanks for the contribution 👍

fabiensc0ville:
To give a bit more a design accessibility/design context here @allmarkedup , you are right to believe paragraphs written in English have maximum legibility at [around 75 characters](https://baymard.com/blog/line-length-readability). The further you get from those numbers, the slower your reading is going to get.

Of course, several parameters have to be considered, such as the font size, weight, the context of reading, etc. But as a general rule of thumb, a lot of research has been made around this topic and there isn't much controversy about it.

Regarding the implementation in Lookbook, as long as we have the choice to constrain the line-width to our preference, I guess it's ok :)

PR #116, opened 2022-07-26
Updating input params to accept date related inputs

jonrohan:
This PR adds [date](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) and [datetime-local](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local) as `@param` input options.

These are both types of `<input>` tags so didn't think it was necessary to add new erb files for the previews.

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,6 +341,8 @@ The following **input field types** are available for use:
 
 ```ruby
 @param <name> text
+@param <name> date
+@param <name> datetime-local
 @param <name> email
 @param <name> number
 @param <name> url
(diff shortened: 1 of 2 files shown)


jonrohan:
That makes sense to me, I was `DateTime.parse(datetime)` in my preview immediately with the data returned.

jonrohan:
PR updated, let me know if I missed anything

allmarkedup:
@jonrohan just given it a spin and it's working great 👍 I'll merge this in now and ship it in the next release.

Many thanks for the contribution, much appreciated!

PR #120, opened 2022-08-05
Use correct cursor for vertical resize

thutterer:
**Before:** ↔️ (same as horizontal resize)
**After:** ↕️


allmarkedup:
@thutterer really sorry for the unbelievably slow response - I've just gotten back from holiday and I'm just catching up with everything now.

PR looks great - good catch! Will merge it in now and included it in the next release. Many thanks :-)

PR #121, opened 2022-08-09
Use prefixed path helper

bradparker:
Sorry, this is similar to the other PR we submitted! We were only
rendering one page so didn't run into this until now 😅

Hopefully this is the right move to keep the 'page_path' helper working
in page content while also supporting disabling the
`config.action_controller.include_all_helpers` Rails configuration
option.

Thanks again, we continue to love using Lookbook. It's been really ace 
linking directly to great looking component previews when talking about
new UIs :)

### DIFF
--- a/app/views/lookbook/pages/show.html.erb
+++ b/app/views/lookbook/pages/show.html.erb
@@ -8,7 +8,7 @@
     <div id="page-banner-<%= @page.id %>" class="flex items-center ml-auto pr-2">
       <% if @previous_page %>
         <a ref="header-previous-page"
-          href="<%= page_path @previous_page.lookup_path %>"
+          href="<%= lookbook_page_path @previous_page.lookup_path %>"
           x-tooltip.theme.lookbook="`Previous page`"
           class="flex-none p-1">
           <%= icon "chevron-left", size: 4, class: "hover:text-indigo-800" %>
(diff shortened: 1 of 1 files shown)


allmarkedup:
Hey @bradparker I'm so sorry for taking so long to reply! I've been away on holiday and expected to have WIFI but it turns out not to have been the case 🙄

I very much appreciate the PR but I've taken a look at this and the issue is actually that I've missed manually including the page helper module into the controller that renders the page views (as opposed to the controller that renders the page content itself, where it _is_ correctly included).

So I'm going to close this PR, fix the upstream issue and get it out in the next release.

Thanks for bringing it to my attention and apologies again for being so slow to get back to you.

bradparker:
No worries! Holidays are made better by a lack of WiFi, I think. 

> I very much appreciate the PR but I've taken a look at this and the issue is actually that I've missed manually including the page helper module into the controller that renders the page views (as opposed to the controller that renders the page content itself, where it is correctly included).

No worries. Much better idea to have the helper included in both controllers 👍

Don't worry about speed in getting back, it's all good from our end. We just amend our fork with what we need, point our bundle at that rev and submit a PR. Always nice to be back on the main repo but never urgent.

Thanks again :)

allmarkedup:
Ok great, in case you haven't updated the latest release should be all fixed for you.

Anything else just let me know! :-)

PR #126, opened 2022-08-31
Disable CSP in development environment

thutterer:
Rails applications that use the their Content Security Policy also in the development environment are likely to block and break Lookbook. When that happens, all the user sees is a blank white page, and some errors in the browser console :eyes: 

This change disables any CSP in the Lookbook controllers, but only in development. Production environments are not impacted.

### DIFF
--- a/app/controllers/lookbook/application_controller.rb
+++ b/app/controllers/lookbook/application_controller.rb
@@ -1,5 +1,6 @@
 module Lookbook
   class ApplicationController < ActionController::Base
+    content_security_policy false, if: -> { Rails.env.development? }
     protect_from_forgery with: :exception
 
     helper Lookbook::ApplicationHelper


pmackay:
Could it be possible for this to be configurable?

The hardest part of integrating Lookbook was getting around CSP, which I've just worked through. I disabled it for all Lookbook controllers for all environments, because CSP is not a concern for lookbook pages. Appreciate that may not be the case for others using CSP, hence the question :)

thutterer:
I thought for this PR it's best to not get into the topic of "lookbook in production" but keep it small and focused. This PR essentially just makes explicit what was always sort of the assumption, please correct me if I'm wrong, @allmarkedup.

All it does is making sure that, in development mode, no CSP gets in the way.

Solving CSP for production, in a configurable manner, is something for a later day, I guess :)

allmarkedup:
Thanks for this @thutterer! I have to admit I know very little about CSPs so it's really good to get some contributions in this area.

Some sort of configuration option(s) could certainly be good in the future but I agree that explicitly disabling CSP here is a good first step. @pmackay what would have been useful to you to help with your integration? Just a straight `true`/`false` option for disabling CSP in all of Lookbook's controllers? Or something a bit more fine-grained?

I think I'll merge this now as it is but I'd love to learn a bit more about people's CSP requirements and how best to support them where possible.

PR #136, opened 2022-09-06
fix: Conditionally call `content_security_policy`

DRBragg:
The `content_security_policy` method isn't available in Rails <= 5.1 so using lookbook >= 0.9.7 fails on older version of Rails. This PR checks that `content_security_policy` is available before trying to call it. This _should_ ensure compatibility with older versions of Rails 🤞 I'll open another PR or an issue if I find anything else.

### DIFF
--- a/app/controllers/lookbook/application_controller.rb
+++ b/app/controllers/lookbook/application_controller.rb
@@ -1,6 +1,9 @@
 module Lookbook
   class ApplicationController < ActionController::Base
-    content_security_policy false, if: -> { Rails.env.development? }
+    if respond_to?(:content_security_policy)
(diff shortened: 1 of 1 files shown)


DRBragg:
@allmarkedup Not sure what's up with the failing GH action. Doesn't appear to be my code but if I did something wrong let me know and I'll fix it.

allmarkedup:
Hey @DRBragg, thanks for this - I didn't realise that wasn't available in older versions of Rails. I'll get it merged in and also update the 0.9.7 branch when I have a chance so that pre-1.0 versions maintain compatibility too.

And don't worry about the failing GH action - I've been struggling to get them working correctly so have removed them for now until I can get them sorted out.

Thanks again as always :-)

PR #137, opened 2022-09-06
Updating docs site for date and datetime-local input options

jonrohan:
Congrats on the 1.0 launch!

I noticed the docs site was missing the input options added in https://github.com/allmarkedup/lookbook/pull/116

I'm adding the param options here


allmarkedup:
Thanks @jonrohan! And not sure how those input options didn't get copied across... must have pulled them from before the PR was merged. Appreciate you adding them back in again :-)

PR #139, opened 2022-09-07
chore: Add Within3 to list of companies using lookbook

DRBragg:
Adding [Within3](https://within3.com/) to the **Who uses Lookbook?** section of the docs. Thanks again for all your work on this!


allmarkedup:
Awesome :-) Thanks @DRBragg!

PR #146, opened 2022-09-13
Add config option to disable action_view annotations

steves:
Rails allows you to set a config option named `config.action_view.annotate_rendered_view_with_filenames` which will wrap rendered templates in HTML comments indicating the file system path of the file being rendered. This can be useful in dev environments to make it easier to track down where parts of the page are coming from. However it can create a bit of noise when viewing the HTML tab of a component in Lookbook. Below is an example of setting this value to `true` in the workbench app.

Screen Shot 2022-09-13 at 12 07 03 PM

This PR introduces a new Lookbook config setting to generate this source without these annotations in the event the setting is enabled. Setting `config.lookbook.disable_action_view_annotations = true` will generate the source without them.

Screen Shot 2022-09-13 at 12 16 39 PM

I made this a configuration setting in the event there are folks who want to leave these annotations in place. I'm defaulting it to `false` because that matches the existing functionality so it feels less disruptive. I'm happy to change the default to `true` if you feel otherwise though.

The `action_view.annotate_rendered_view_with_filenames` config option was first added to Rails in 6.1.0. I'm not sure the best way to test this in versions prior to that to ensure it works or which versions we want to ensure Lookbook is compatible with.

### DIFF
--- a/docs/src/_data/config_options.yml
+++ b/docs/src/_data/config_options.yml
@@ -168,7 +168,7 @@
 - name: preview_disable_action_view_annotations
   group: string
   type: Boolean
-  default: false
+  default: true
   example: "config.lookbook.preview_disable_action_view_annotations = true"
   description: "Turns off action view filename annotations when generating rendered component source."
 
--- a/docs/src/_data/config_options.yml
+++ b/docs/src/_data/config_options.yml
@@ -165,11 +165,11 @@
   example: config.lookbook.parser_registry_path = "path/to/temporary/storage"
   description: The directory to write the (temporary) parser registry file to.
 
-- name: disable_action_view_annotations
+- name: preview_disable_action_view_annotations
   group: string
   type: Boolean
   default: false
-  example: "config.lookbook.disable_action_view_annotations = true"
+  example: "config.lookbook.preview_disable_action_view_annotations = true"
   description: "Turns off action view filename annotations when generating rendered component source."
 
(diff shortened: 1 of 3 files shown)


allmarkedup:
@steves this is really great! Many thanks for this.

Just one minor thing - I've been trying to 'namespace' related config options via a prefix on the name. So as this option is related to previews I think it should probably be called `preview_disable_action_view_annotations` - it's a bit of a mouthful but follows the convention of the other preview related options. What do you think? Would you be able to update the PR if you agree?

Also my gut feeling with this is to set the default value to `true`, even if that is a change in the current default behaviour. Looking at the screenshot you provided, the annotations add a lot of noise and if they can be easily turned back on I think stripping them out is probably a better default for new users. Do you agree?

Let me know what you think and then I'll get this merged in when ready.

steves:
@allmarkedup Thanks for the feedback. Those changes make sense to me. I'll make them today and get the PR updated.

@DRBragg Thanks for the input. I was thinking of adding a `respond_to?` check to be on the safe side. I tend to prefer checks like that over version numbers as I find them more self documenting.

allmarkedup:
@steves @DRBragg yep the `respond_to?` check works for me, nice one 👍

@steves thanks for making those changes, all looks good to me now so I'm going to merge this in and will include it in the next release. Really appreciate the contribution!

PR #153, opened 2022-09-19
Add base auth to lookbook

Nookbe:
Background:
To give our designers access to lookbook, we activated lookbook in our production environment.
For security reasons, we at least want to provide a base_auth so that not everyone has access to our component library in lookbook.

What was done:
- I added a before_action to the application_controller
- I added a config so that base_auth is deactivated by default

### DIFF
--- a/app/controllers/lookbook/application_controller.rb
+++ b/app/controllers/lookbook/application_controller.rb
@@ -1,5 +1,7 @@
 module Lookbook
   class ApplicationController < ActionController::Base
+    before_action :base_auth, if: config.base_auth && Rails.env.production?
+
     if respond_to?(:content_security_policy)
       content_security_policy false, if: -> { Rails.env.development? }
     end
@@ -27,6 +29,10 @@ module Lookbook
     end
 
     protected
+
+    def base_auth
+      http_basic_authenticate_with name: ENV['LOOKBOOK_BASE_AUTH_NAME'], password: ENV['LOOKBOOK_BASE_AUTH_PASSWORD']
+    end
 
     def generate_theme_overrides
       @theme_overrides ||= Lookbook.theme.to_css
--- a/lib/lookbook/config.rb
+++ b/lib/lookbook/config.rb
@@ -40,6 +40,8 @@ module Lookbook
         ui_theme_overrides: {},
         ui_favicon: true,
 
+        base_auth: false,
+
         hooks: {
           after_initialize: [],
           before_exit: [],
@@ -58,7 +60,7 @@ module Lookbook
             hotkey: "v",
             panel_classes: "overflow-hidden",
             padded: false,
-            system: true 
+            system: true
           },
           output: {
             pane: :main,


app/controllers/lookbook/application_controller.rb
```diff
@@ -1,5 +1,7 @@
 module Lookbook
   class ApplicationController < ActionController::Base
+    before_action :base_auth, if: config.base_auth && Rails.env.production?
```
allmarkedup, line 3: Just looking at this, I wonder if checking for `Rails.env.production?` is needed here - it feels like the end user could do that when setting the `config.base_auth` setting value if that is the behaviour they wanted. For example:

```rb
# config/application.rb
config.lookbook.base_auth = Rails.env.production?
```

So Lookbook itself is not making any additional assumptions past the config value check. What do you think?

allmarkedup (commented):
Hey @Nookbe many thanks for this! I agree this is a useful feature to have for those that choose to expose their Lookbook instance in production.

Just one minor comment on the `Rails.env.production?` check - let me know what you think. But otherwise I like this and will be happy to  get it merged in and added to the next release. Thanks again!

allmarkedup:
@Nookbe hmm the CI are currently failing - I'll look into that when I have a chance

allmarkedup:
@Nookbe this is also not working for me locally, just trying to figure out why not!

allmarkedup:
Ok so to get this to work for me I needed to make two changes:

```rb
before_action :base_auth, if: -> { Lookbook.config.base_auth && Rails.env.production? }
```
and

```rb
def base_auth
      self.class.http_basic_authenticate_with name: ENV['LOOKBOOK_BASE_AUTH_NAME'], password: ENV['LOOKBOOK_BASE_AUTH_PASSWORD']
    end
```

I've no idea why I need to invoke `http_basic_authenticate_with` as a class method here? But it seems to work. I'd be curious to understand why this is working for you and not for me though...

allmarkedup:
@henrikbjorn Thanks for your input on this. Given that there are ways to implement this outside of Lookbook (as per your examples) I think I'll close this now. 

@Nookbe thanks again for opening the PR and if you feel strongly that there are reasons this should be integrated into Lookbook that are not covered by one of the suggestions by @henrikbjorn then let me know and I'd be happy to reconsider.

henrik:
I think it makes sense to add to Lookbook.

The "authenticate" method is part of Devise, so doesn't work in plain Rails, and I believe it sets up a routing constraint which by some is [considered harmful](https://anadea.info/blog/rails-authentication-routing-constraints-considered-harmful).

I tried wrapping Lookbook in another Rack app as mentioned in the linked blog post, but that did not pass through the Lookbook routes, so things didn't work.

I may keep exploring this and will share anything I come up with, but I think Lookbook should offer either something built in or document a pattern for it.

henrik:
What I've come up with so far is basically what this PR does.

For example, stick this in `config/initializers/lookbook.rb`:

``` ruby
Rails.application.config.after_initialize do
  class Lookbook::ApplicationController
    http_basic_authenticate_with name: "secret", password: "also secret"
  end
end
```

PR #161, opened 2022-09-22
fix: Ensure key is a string for `start_with?`

DRBragg:
In `lib/lookbook/theme.rb:44` it checks `key.start_with?("favicon")`. Unfortunately, `key` is a symbol and support for `start_with?` for symbols wasn't added until Ruby 2.7

Sorry to be constantly opening PR's to support older versions of Ruby but Lookbook has be super awesome for us and we'd love to keep using it! Hopefully after we update I can start contributing real features and not just patches ❤️ 

I also did a minor syntax change from `select` to `reject` for filtering since I think it's a little clearer to reject keys that start with "faviocon" rather than select keys that dont start with favicon.  But this might just be my preference so I can roll it back if you want.

Thanks for all you do!

### DIFF
--- a/lib/lookbook/theme.rb
+++ b/lib/lookbook/theme.rb
@@ -41,7 +41,7 @@ module Lookbook
       return @css unless @css.nil?
       @css ||= if @overrides.present?
         styles = [":root {"]
-        styles << @overrides.select { |key| !key.start_with?("favicon") }.map do |key, value|
+        styles << @overrides.reject { |key| key.to_s.start_with?("favicon") }.map do |key, value|
           "  --lookbook-#{key.to_s.underscore.tr("_", "-")}: #{value};"
         end
         styles.push "}"


allmarkedup:
@DRBragg please don't apologise - you are doing great work here! I'm sure there are plenty of other people who really appreciate you keeping Lookbook working with older Ruby versions. What version are you working with again? It was 2.5 wasn't it?

I really should get the Github Actions CI configured to test against some older Ruby versions too so I can hopefully catch some of this stuff a bit earlier - I ran into a few problems getting it working before but it should be perfectly possible.

And totally agree with your `select/reject` reasoning so very happy to keep that in.

Thanks again - will merge this now.

DRBragg:
> @DRBragg please don't apologise - you are doing great work here! I'm sure there are plenty of other people who really appreciate you keeping Lookbook working with older Ruby versions. What version are you working with again? It was 2.5 wasn't it?
> 
> I really should get the Github Actions CI configured to test against some older Ruby versions too so I can hopefully catch some of this stuff a bit earlier - I ran into a few problems getting it working before but it should be perfectly possible.
> 
> And totally agree with your `select/reject` reasoning so very happy to keep that in.
> 
> Thanks again - will merge this now.

Yea, 2.5 for now. Luckily I feel a little better about this PR since it also keeps Lookbook compatible with Ruby 2.6

Thanks for the quick response, you rock!

PR #175, opened 2022-09-29
Add config option to make rails logger optional

steves:
Our local development environment uses a verbose logging level and the Lookbook logs are adding quite a bit of extra stuff to them. I'm proposing a new configuration option that would let us tell Lookbook not to use the Rails logger even though we're running inside a Rails app. This will give us a bit more control over our logs.

I've defaulted the value to `true` so it should match the existing behavior.

### DIFF
--- a/docs/src/_data/config_options.yml
+++ b/docs/src/_data/config_options.yml
@@ -15,6 +15,13 @@
     _Only applicable if no logger has been set in the parent application_, otherwise Lookbook
     will use that logger and the log level value that has been set on it.
 
+- name: log_use_rails_logger
+  group: debugging
+  type: Boolean
+  default: true
+  example: config.lookbook.log_use_rails_logger = true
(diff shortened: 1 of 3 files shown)


allmarkedup:
Hey @steves, thanks for this! I feel like I've never really got the logging story right with Lookbook (lots of people have lots of different setups/requirements which makes it hard to find a good fit for everyone) so it's definitely good hear how it's being used and to provide some options like this.

I'm actually wondering now if it would be better for Lookbook to log everything to `stdout` by default. And then let people  provide a different logger via a config option if they want one - either pass in the Rails one in or a custom one. Or disable it altogether.

What do you think? Would that be a better solution than the current one for you?

I think I'll merge this in now anyway as it's good to have this option for the current setup but I'd be interested to hear your opinion on the above. It would probably only change in the next major release as it would break with existing behaviour but I wonder if that would be a better default for people.

Really appreciate you taking the time to contribute back to Lookbook again 👍

Issue #87, opened 2022-05-05
Component theme switcher 


allmarkedup:
Some component libraries (like [Primer](https://github.com/primer/view_components)) offer multiple themes for their components.

To help with previewing components in themed libraries, it should be pretty straightforward to add a 'component theme' dropdown that allows selecting a theme from a configurable list. The name of the selected theme is then passed down to the preview template so that different stylesheets (or classes) can be negotiated in whichever manner works best for the parent project.

Primer currently have one such example of a switcher like this [in their Storybook](https://primer.style/view-components/stories/?path=/story/primer-alpha-border-box-header--default).

_(originally suggested by @joelhawksley over email)_

jalyna:
I just stumbled across this today. This would really be a great addition. We probably also would like to use the concept for switching languages (currently we use params per component).

allmarkedup:
@jalyna that's good to hear - I'm planning on implementing it in as generic a way as possible so that it can be used for themes or languages or any other kind of persistent user-controlled setting. I'll try to let you know when there is something to test out :-)

jonrohan:
👋🏻 Dropping by on this to hopefully had some insight on where we are with support for it and what would be useful from the library. Maybe a more general plugin framework would work here?

Our color modes work with a set of data attributes to determine theme, etc. I added support for these in our previews here https://github.com/primer/view_components/pull/1451

I think the bit we're missing is some way to add a param option to the top toolbar that would be persisted across previews. 
<img width="1147" alt="image" src="https://user-images.githubusercontent.com/54012/195217669-861ed654-11ae-4fca-8594-5f0f42455671.png">

This would allow us to take that data and apply it on the preview templates.

allmarkedup:
Hey @jonrohan @jalyna - I thought I'd have a bit of a bash on this and it turned out to be reasonably straightforward to implement something that I think should work for your use cases.

Rather than introducing an entirely new concept I've instead extended the display options feature to provide 'dynamic display options' that you can configure as you like to suit your requirements.

There is more detail [in the PR](https://github.com/allmarkedup/lookbook/pull/194) and some [first-pass documentation here](https://deploy-preview-194--lookbook-docs.netlify.app/guide/previews/display/#dynamic-display-options) - I'd be very interested in getting your feedback if you were able to take a look at it!


If you wanted to give it a test you can install directly from the `dynamic-display-options` branch as follows:

```ruby
[...]

PR #194, opened 2022-10-12
Dynamic, persistent display options

allmarkedup:
This PR addresses issue #87 in a generic fashion by adding support for 'dynamic' [display option values](https://lookbook.build/guide/previews/display/) that can be changed by the user in the Lookbook UI and will persist across component previews.

Dynamic display options are defined in the `preview_display_options` configuration by providing an array of possible values:

```ruby
# config/application.rb
config.lookbook.preview_display_options = {
  theme: ["light", "dark"]
}
```

To prevent overlap issues on smaller screen sizes, if more than one dynamic display option is defined they are rendered in a dropdown rather than being directly accessible from the toolbar (although this could be revisited in the future to work out a nicer solution).

CleanShot 2022-10-12 at 23 22 29


jonrohan:
This is fantastic! 👏🏻 

I'm testing it out here https://github.com/primer/view_components/pull/1490/, with our preview app deploying to http://primer-view-components-preview-1490.eastus.azurecontainer.io/view-components/lookbook/inspect/primer/beta/flash/default It was super easy to get working.

There was a couple of comments/questions when I was able to get this working.

1. Is there some way to sync the url with a the dynamic param option? The designers on the team want to be able to change the dropdown to copy the url to share with reviewers. As a way of dropping someone directly into one of the options.
2. It would be nice for the configuration to also support a hash so that select options can have different values and labels.

```rb
config.lookbook.preview_display_options = {
  theme: {
    "light": "Light theme", 
    "dark": "Dark theme"
  }
}
```

allmarkedup:
Hey @jonrohan - thanks for testing it out and for the suggestions.

With regards to sharing the link to the preview, rather than having the dropdown update a param value in the URL when changed (which would be tricky to ensure was persisted across navigation events etc) how about the following:

- If the URL has a `theme` query string param in it, set the option to that value and persist across pages (actually may need to be namespaced as `display-theme` or something to avoid clashes, but something along those lines).
- have a 'share' button next to the display option inputs that when clicked copies a URL to the clipboard that includes the `theme` query string param, plus any others

Obviously it would also be possible to update the URL when the dropdown is changed. But as navigating around may drop that param from the URL it might not be obvious that (after navigation) copying the URL from the browser URL bar might not include the `theme` param in this case. So it's maybe better not to set up that expectation in the first place and instead only rely on a copyable 'share' link instead. What do you think - does that make sense?

And with regards to the labels/values - it sounds like you have sorted that now, but an example in the docs would probably be good to make that explicit. I'll add that in. You can actually already use a hash in the format of:

```rb
config.lookbook.preview_display_options = {
  theme: {
     choices: ["light", "dark"]
  }
}
```

Which only exists as a potential way to potentially expand the functionality in the future by allowing other options to be passed in (perhaps for different input types etc). But good to keep it simple for now I think.

Let me know what you think about the URL handing suggestion and I'll look to get something along those lines added in as soon as I can.

jonrohan:
Ah yeah, I see the predicament with the url params clashing.

What if you kept navigating and changing like it is now, where it doesn't update the url. But then allow a parameter to be passed in that would update the cookie state and select menu state of the dynamic parameter. something like `url?dynamic=theme:dark;lang:en`. 

And include a copy link icon that will bring someone to that page with the same configuration? image

allmarkedup:
@jonrohan yep I think that is pretty much what I was (less clearly and succinctly) trying to suggest too :-) I'll add that in and let you know when it's ready to have a play with but I think that should work 👍

allmarkedup:
@jonrohan I've just pushed some changes to implement what we discussed above. I did end up updating the URL when the select is changed because otherwise if there was already a value for it in the URL then it would get out of sync with the dropdown which obviously isn't good.

The query string will not persist when navigating around however, so the safest way to copy a link to the page is via the new button I added into the toolbar to the left of the display controls.

Let me know how you get on!

allmarkedup:
Hmm I seem to have broken the tests... I think it should all work however, it's just that I changed the way it handled the setting of cookies on the client side.

I will fix the tests tomorrow but I'm going to bed now :-)

jonrohan:
> I will fix the tests tomorrow but I'm going to bed now :-)

No problem, I'll check it out and leave you notes for later.

app/views/lookbook/previews/show.html.erb
```diff
@@ -17,8 +17,24 @@
...
+          <% end %>
+        <% end %>
+      <% end %>
+
+      <% toolbar.section divide: :left, class: "flex-none relative z-10" do %>
         <%= lookbook_render :button_group do |group| %>
+          <% group.button icon: :link,
+            tooltip: "Copy preview URL",
+            copy: true do %>
+            <%= lookbook_inspect_url(@target.lookup_path, request.query_parameters) %>
```
jonrohan, line 35: I would expect that the copy preview url would include the current state of the `static_display_options`. So that if I clicked on copy, it would include `?_display=theme%3Adark` even if that's not currently in the browser url.

allmarkedup: @jonrohan urgh yes it should - that was the whole point of it! Just me not paying enough attention before bed. Will fix that, thanks.

allmarkedup: @jonrohan should be all fixed now 👍

jonrohan (approved):
Everything works great on my end! Thanks for tackling this, it's going to be huge, and I like the approach that gives room for other options.

I did a general scan of the code for typo type mistakes, I can't give a too much in depth review of the system. There were a couple of things that I noticed that might not be anything so feel free to ignore or dismiss.

docs/src/guide/previews/display.md
```diff
@@ -73,14 +71,70 @@ Ruby YAML does not (generally) require quoting of string values. However in some
...
 ```
 
-Globally defined display params will be available to all previews.
-Any preview or example-level `@display` values with the same name will take precedence and override a globally-set one.
\ No newline at end of file
+In the example above, the `default` preview example will set the `theme` display option value to `light` and it will not be possible to choose a different theme whilst viewing that preview.
+
+Overridden values like these **are not persisted** across other previews.
+
+{{toc}}
```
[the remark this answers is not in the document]
allmarkedup: Thanks @jonrohan, that was intentional - it's just to make things more consistent with other docs pages - the `toc` tag can go anywhere but I'm trying to standardise with it just being at the bottom of all pages.

allmarkedup:
@jonrohan many thanks for the review, have added that docs suggestion now but the other changes were intentional although probably shouldn't have been part of the PR - I was just doing a little housekeeping while I was in there!

I'm actually in the middle of trying to do a big refactor of the codebase as when I first built Lookbook I didn't really know much Ruby (I still don't but have learnt a lot since then!) so there is lots to unpick and improve. So I appreciate any and all comments/reviews/suggestions etc :-)

I'm going to merge this in now and will get a new release out in the next day or two. Thanks again for all the help testing it out.

PR #207, opened 2022-10-24
Add ability to customize the folder previews appear in

camertron:
The Primer team would like to customize the folder a preview class lives in, a concept I'm calling the "logical path." Right now, the path of the file on disk is used to create a folder hierarchy. This PR introduces a new `@logical_path` tag that overrides the default path with a custom path of your choosing. For example, let's say I'd like to re-scope the `NavList` previews under foo/bar:

```ruby
# @logical_path primer/foo/bar/nav_list
class NavListPreview < ViewComponent::Preview
  # ...
end
```

Here's how it looks:

image

It works! 🎉

### DIFF
--- a/config/tags.yml
+++ b/config/tags.yml
@@ -27,3 +27,8 @@ shared:
     label: Param
     yard_parser: with_name
     opts: {}
+
+  logical_path:
+    label: Logical path
+    named_args: [logical_path]
+    opts: {}
--- a/docs/src/guide/previews/navigation.md
+++ b/docs/src/guide/previews/navigation.md
@@ -29,6 +29,23 @@ class FooComponentPreview < ViewComponent::Preview
 end
 ```
 
+## Logical Paths
+
+Use the `@logical_path` annotation tag to change the nav folder the preview will appear in.
+
+```ruby
+@logical_path <path>
+```
+
+The `@logical_path` tag can only be used on preview classes:
+
+```ruby
+# @logical_path path/to/my/component
+class FooComponentPreview < ViewComponent::Preview
+  # ...
+end
+```
+
 ## Hiding previews
 
 By default, all previews and examples are shown in the preview navigation.
--- a/lib/lookbook/preview.rb
+++ b/lib/lookbook/preview.rb
@@ -8,7 +8,8 @@ module Lookbook
     def initialize(preview, code_object)
       @preview = preview
       @preview_inspector = SourceInspector.new(code_object, eval_scope: preview_class.new)
-      super(preview_class_path(@preview.name))
+      logical_path = @preview_inspector.tag(:logical_path)&.logical_path
+      super(logical_path || preview_class_path(@preview.name))
     end
 
     def id
--- a/lib/lookbook/stores/tag_store.rb
+++ b/lib/lookbook/stores/tag_store.rb
@@ -39,7 +39,8 @@ module Lookbook
         name: name.to_sym,
         label: opts[:label] || name.to_s.titleize,
         yard_parser: opts[:yard_parser]&.to_sym,
-        opts: opts[:opts].to_h
+        opts: opts[:opts].to_h,
+        named_args: opts[:named_args]
       })
     end
   end
--- a/lib/lookbook/tags.rb
+++ b/lib/lookbook/tags.rb
@@ -6,9 +6,9 @@ module Lookbook
         tag_config = Engine.tags.get_tag(tag_object.tag_name).to_h
         tag_opts = tag_config[:opts].to_h
         Lookbook::Tag.new(tag_object,
-          tag_opts[:named_args],
+          tag_config[:named_args],
           parser: tag_opts[:args_parser],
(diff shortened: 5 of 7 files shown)


allmarkedup:
Hey @camertron, many thanks for this! I like the idea of being able to override the path of a preview a lot. 

However I'd be interested in learning a little more about the use-case for this in your particular situation. Is it just the occasional preview that you are wanting the need to relocate? Or is it more that you want to be able to effectively construct an entire custom navigation structure, independent from the disk paths?

Obviously your implementation could probably serve both use-cases, but I'm just wondering if the latter might be better served by a solution that wouldn't require editing multiple files in order to (for example) change the 'logical path' to one or more entire directories of previews. Some sort of top-down configuration for specifying a tree and what previews to include in it at what locations, for example. But that may not be what you are needing this for at all of course! :-)

Regardless, one slight issue with merging this in is that you've caught me right in the middle of a pretty big refactor of the whole tag parsing system to clean up and rationalise some of the horribleness in there! I've got a [work-in-progress PR up here](https://github.com/allmarkedup/lookbook/pull/208) which is almost ready to go once I've added some more tests.

Are you happy to hang on until I've merged the refactor and then I can come back to this and help get it updated to work with the new system? It should (hopefully) make it implementing new tags like this a bit of a saner process.

camertron:
> Hey @camertron, many thanks for this! I like the idea of being able to override the path of a preview a lot.

Cool, and no problem! From my (limited) exploration, Lookbook appears to be exceptionally well-written 😄 

> However I'd be interested in learning a little more about the use-case for this in your particular situation. Is it just the occasional preview that you are wanting the need to relocate? Or is it more that you want to be able to effectively construct an entire custom navigation structure, independent from the disk paths?

Our main use-case right now is to be able to add a "Features" sub folder for some additional previews that aren't good candidates for the default and playground previews we have in the main folder. For example, we want Alpha/Button but also Alpha/Button/Features.

> I'm just wondering if the latter might be better served by a solution that wouldn't require editing multiple files in order to (for example) change the 'logical path' to one or more entire directories of previews. Some sort of top-down configuration for specifying a tree and what previews to include in it at what locations, for example.

That's an interesting idea, but not what we need at the moment. I also like being able to see the logical path right there in the preview code instead of having to look for it in a separate config file. Seems less potentially surprising maybe?

> Are you happy to hang on until I've merged the refactor and then I can come back to this and help get it updated to work with the new system?

Yes, absolutely no problem! Thanks for your willingness to help :)

allmarkedup:
>  From my (limited) exploration, Lookbook appears to be exceptionally well-written 😄

Whilst I appreciate the vote of confidence @camertron this is my first Ruby project and as it all grew up pretty organically there is some very convoluted code in there 😂 But I'm working through a lot of refactoring now which should hopefully make it easier for people to contribute and understand what is going on :-)

> Our main use-case right now is to be able to add a "Features" sub folder for some additional previews that aren't good candidates for the default and playground previews we have in the main folder. For example, we want Alpha/Button but also Alpha/Button/Features.

Okay cool, that makes perfect sense. I just wanted to check that you weren't after a way to customise the entire nav tree - I'd love to have a clever, low-maintenance way to support that but that is a bigger project for another day!

I totally agree about it being nice to have the logical path right there in the preview code too. The more I think about this feature the more I like it actually.

So, with regards to getting this merged in, I've now merged my tag parsing refactor work into main which means that your changes here will now not work as expected. I'm very happy to take what you've done here and re-implement it myself if you are happy for me to do that? It would be a good test for myself of how easy it is to add another system tag after my refactoring.

The only issue with that is that you will lose attribution for your PR work above - unless you want to give me access to push to your fork and then I can commit changes to this PR directly. Whatever you feel happiest with.

I can also pair with you at some point to talk through the changes if you like but it may be quicker for me just to re-implement it!

Let me know how you'd like to proceed and thanks for taking the time to give me a little more context above.

camertron:
> I totally agree about it being nice to have the logical path right there in the preview code too. The more I think about this feature the more I like it actually.

😄 

> So, with regards to getting this merged in, I've now merged my tag parsing refactor work into main which means that your changes here will now not work as expected. I'm very happy to take what you've done here and re-implement it myself if you are happy for me to do that? It would be a good test for myself of how easy it is to add another system tag after my refactoring.

If you've got the time and the inclination, I'd be more than happy if you re-implemented it!

> The only issue with that is that you will lose attribution for your PR work above - unless you want to give me access to push to your fork and then I can commit changes to this PR directly. Whatever you feel happiest with.

That's really nice of you to consider, but it's no problem - I don't need attribution 😄

allmarkedup:
@camertron I've just opened a PR here with the re-implemented `@logical_path` tag - if you could give it a quick look over at some point it would be much appreciated.

I've made a small change so that the `@logical_path` value should the **directory** path rather than the path to the preview itself. So that naming and labelling of the previews all still happens consistently. How do you feel about that?

If you are happy with it I'll get it merged in and out in the release.

PR #215, opened 2022-10-29
Add @logical_path tag

allmarkedup:
As per discussion in https://github.com/ViewComponent/lookbook/pull/207 this adds a new tag `@logical_path` which can be used to relocate previews within the nav tree.

The `@logical_path` value should be the path to the nav directory that the preview should be displayed in. The provided path does not have to really exist in the app itself, it is just a virtual construct.

```rb
# @logical_path foo/bar
class Elements::AvatarComponentPreview < ViewComponent::Preview  
 #...
end
```

CleanShot 2022-10-29 at 01 04 15

### DIFF
--- a/config/tags.yml
+++ b/config/tags.yml
@@ -26,3 +26,7 @@ shared:
   param:
     label: Param
     opts: {}
+
+  logical_path:
+    label: Logical Path
+    opts: {}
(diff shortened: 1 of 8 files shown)


camertron (approved):
This looks great, thank you! 🎉

PR #222, opened 2022-11-04
More theme overrides options

mdrbohlav:
- Add `--lookbook-viewport-handle-hover` CSS variable to be able to override it but keep the `--lookbook-draggable-hint` which is used on multiple places.
- Remove hardcoded white background from `Params::Editor::Component`.
- Use `--border-lookbook-divider` for border colors in the `Params::Field::Component`.
- Set `Viewport::Component` iframe background color to be the same as the page's background color. Makes the possible custom designs look much better (e.g. with dark colors).


allmarkedup:
@mdrbohlav Really appreciate your eye for detail here. This all looks good and I'm going to merge it in now.

The one thing I'm going to add a follow-up commit on is to add a CSS variable for the `Params::Editor::Component` background colour (set to white) as that was a specific design decision for that table. But it definitely makes sense to make that configurable along with everything else.

Many thanks again :-)

allmarkedup:
This has been merged manually as there were some merge conflicts to resolve 👍

Issue #233, opened 2022-11-10
Parameter controls not working


jacob-carlborg-apoex:
### Describe the bug

I've integrated Lookbook to our application but some of the parameter controls are not working. The once I've tested which are not working are: `toggle` and `select`. Nothing happens when I click on a toggle or a select box.

### To Reproduce

1. Go to the Lookbook page of the application
2. Select a component
3. Click on the `Params` tab 
4. Click on the toggle for a parameter
5. Nothing happens and no errors in the browser console

### Expected behavior

I expect the toggle to change state and the preview to be updated.

### Version numbers 

* Lookbook: 1.3.3
* ViewComponent: 2.74.1
* Rails: 6.1.5.1
* Ruby: 2.7.2p137 (2020-10-01 revision 5445e04352) [arm64-darwin20]
* view_component-contrib: 0.1.1
* Safari: 16.1 (18614.2.9.1.12)
* Chrome: 107.0.5304.87 (Official Build) (arm64)

### Additional context

It seems to be some issue with the JavaScript, but there are no errors in the browser console. The tree that lists the components are working properly so the JavaScript is not completely broken. If I change the parameter type to `text` everything works fine and the preview is automatically updated when I change the parameter value.

I've also tested to generate a new Rails application with ViewComponent and Lookbook and then everything works as expected. So there's something weird with the setup/configuration in our application. I've tried to remove gems from our application until only the standard Rails gems (those that are included when generating a new Rails application) are left plus ViewComponent and Lookbook, more or less. It didn't help.

I've included the debug data from Lookbook below. Note, `config.components_path` is not correct. We have our components in `app/frontend/components` and `app/frontend/pages`.

```json
{
  "version": "1.3.3",
  "env": "development",
  "config": {
    "project_name": "Lookbook",
    "log_level": 2,
    "log_use_rails_logger": true,
    "auto_refresh": true,
    "components_path": "app/components",
    "page_controller": "Lookbook::PageController",
    "page_route": "pages",
    "page_paths": [
      "test/components/docs"
    ],
    "page_extensions": [
      "html.*",
      "md.*"
    ],
    "page_options": {
    },
    "markdown_options": {
      "tables": true,
      "fenced_code_blocks": true,
      "disable_indented_code_blocks": true,
      "strikethrough": true,
      "highlight": true,
      "with_toc_data": true,
      "lax_spacing": true
    },
    "highlighter_options": {
      "theme": "github",
      "dark": false
    },
[...]

PR #243, opened 2022-11-24
Fix toggle input attribute-encoding issue  

allmarkedup:
Fixes an issue (#233) where the toggle input was completely unresponsive on some setups due to some over-zealous attribute encoding effectively disabling the AlpineJS `@click` handler.

Many thanks to @jacob-carlborg-apoex for his help in tracking this one down 👍

### DIFF
--- a/app/components/lookbook/params/field/component.rb
+++ b/app/components/lookbook/params/field/component.rb
@@ -34,7 +34,7 @@ module Lookbook
           value: param.value,
           value_type: param.value_type,
           value_default: param.value_default,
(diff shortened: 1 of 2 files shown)


jacob-carlborg-apoex:
@allmarkedup I had the same problem with select boxes as well. If there are other parameter types that depend on JavaScript I would assume they have the same problem as well.

allmarkedup:
@jacob-carlborg-apoex I've not yet been able to replicate the same issue on select boxes - I'll merge this in anyway to go out in the next release and take another look at it over the weekend.

PR #253, opened 2022-12-04
Performance improvements

kirillplatonov:
We use lookbook in production as a documentation website for [Polaris ViewComponents](https://polarisviewcomponents.org). The loading time on production was pretty slow (2-3 seconds for the initial inspector load). After diving into the problem I noticed that the main source of the issue is sidebar generation. It was called 3 times for each request and generation itself takes some time for a large preview list. To optimize this I made two changes:

1. Added memoization for tree generation in the Nav component. It's called 3 times per request so it's a big win in development mode.
2. I added caching for the whole lookbook sidebar. It makes loading fast on production as we don't need to regenerate the tree each time.

### DIFF
--- a/app/components/lookbook/nav/component.rb
+++ b/app/components/lookbook/nav/component.rb
@@ -12,7 +12,7 @@ module Lookbook
     end
 
     def items
-      tree.map do |node|
+      @items ||= tree.map do |node|
         item_class = (node.type == :directory) ? Nav::Directory::Component : Nav::Entity::Component
         lookbook_render item_class.new node, nav_id: id
       end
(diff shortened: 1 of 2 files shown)


allmarkedup:
@kirillplatonov thank you so much for this! I have been pretty busy but will try to give it a test and merge it in in the next few days. Really appreciate the contribution 👍

allmarkedup:
@kirillplatonov Given this a spin now and it's looking good, many thanks again 👍

Merging now and will get it out in the next release soon.

kirillplatonov:
Thanks, @allmarkedup. Awesome work on this gem! It keeps getting better with every release 👏

allmarkedup:
@kirillplatonov Ah thank you, I'm just trying to work out all the rough edges bit by bit :-)

I'm aware the experience of working with large component libraries is not optimal yet so contributions like this are much appreciated. I'm working on some changes which should hopefully see some other performance improvements to the development experience too. One step at a time we'll get there!

PR #266, opened 2022-12-12
chore: Require Ruby version 2.7 or higher

DRBragg:
Lookbook isn't compatible with apps running Ruby < 2.7 so we should update the gemspec to reflect that. At present you can install lookbook in an app running <2.7 without any bundler errors but will cause booting the app to fail. Updating the gemspec will stop people with unsupported ruby versions from installing in the first place.

### DIFF
--- a/lookbook.gemspec
+++ b/lookbook.gemspec
@@ -14,6 +14,8 @@ Gem::Specification.new do |spec|
   spec.files = Dir["{app,config,lib,public}/**/*", "LICENSE.txt", "README.md"]
   spec.require_paths = ["lib"]
 
+  spec.required_ruby_version = ">= 2.7.0"
+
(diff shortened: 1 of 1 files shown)


allmarkedup:
Hey @DRBragg thanks for this, and once again the thanks come along with apologies that I've not made any more headway with improving Lookbook's backwards compatibility 🙄

Definitely better to not mislead people who are using older Ruby versions though so this is great. Will merge it in now, cheers.

DRBragg:
It's understandable really. 2.6 was EOLed like 8 months ago. I def don't fault you for only supporting currently supported Rubies. This pin is just the right call for any gem imo.

It's probably also a good idea to pin `actioncable` and `activemodel` to the same version as `railties` also but less important.

PR #284, opened 2023-01-11
Fix path to input types documentation

stevegeek:
A small fix to the documentation site where a path was accidentally set as an absolute path to localhost


allmarkedup:
Great spot - thanks for this @stevegeek :-)

PR #288, opened 2023-01-16
Remove redundant `Utils` include

paracycle:
The `include Utils` line in `PageHelper` seems to be a leftover from a refactor. The `Utils` module has no methods of its own, only its singleton class has methods. So including `Utils` into any class will not have any effect, which makes this include redundant.


allmarkedup:
Thanks @paracycle - you are correct about being left over from a refactor. Appreciate you taking the time to open the PR and get things cleaned up 👍

PR #289, opened 2023-01-17
Link to official Rails docs instead of apidock

TastyPi:
apidock is out-of-date, it only goes up to Rails 6.1.3.1 which was released in 2021, much better to link directly to the latest official Rails docs.


allmarkedup:
@TastyPi yes much better, thank you! 🙏

PR #291, opened 2023-01-19
Add `title` to `iframe` in `Viewport::Component`

TastyPi:
I'm working on a panel that runs axe tests on the preview. To accomplish this, I pass the `iframe` element to `axe.run`. Currently this always produces a violation because the iframe is missing a `title` (see https://dequeuniversity.com/rules/axe/4.6/frame-title). This will let me avoid having to filter out the violation and it improves Lookbook's accessibility (which I gotta say isn't great at the moment, a lot of unlabelled buttons).


allmarkedup:
@TastyPi yes I'm afraid the state of Lookbook's accessibility is a bit embarrassing at the moment 😞 It's on the list of things to improve so any help like this is much appreciated, thank you!

PR #300, opened 2023-01-24
Add JSON endpoint for listing previews and examples

joelhawksley:
This change introduces a JSON endpoint that returns a machine-readable representation of the previews and examples served by Lookbook.

I'm introducing this feature so that Lookbook deployments can be crawled by our automated tooling at GitHub.

### DIFF
--- a/docs/src/api/json_endpoints.md
+++ b/docs/src/api/json_endpoints.md
@@ -1,0 +1,19 @@
+---
+layout: default
+title: JSON Endpoints
+---
+
+Lookbook currently implements a single JSON endpoint at `/previews.json`, returning previews and their examples:
+
+```ruby
+[{
+  "name" => "annotated",
+  "examples" =>
+    [
+      {
+        "name" => "default",
+        "inspect_path" => "/lookbook/inspect/foo/bar/annotated/default"
+      }
+    ]
+}]
+```
--- a/spec/requests/application_spec.rb
+++ b/spec/requests/application_spec.rb
@@ -75,4 +75,18 @@ RSpec.describe "application", type: :request do
       end
     end
   end
+
+  context "JSON endpoint" do
+    it "returns a JSON list of previews with example inspect paths" do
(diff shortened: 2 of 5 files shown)


Spone:
What do you think about creating a `PreviewsController#index` action for this, instead of reusing the home?

I find it a little weird to have an action that displays the landing page when requesting HTML, but queries previews when requesting JSON format.

Also, if the route is `/previews.json`, it's immediately clear what the content will be.

joelhawksley:
@Spone great idea. I've updated this PR accordingly!

docs/src/api/json_endpoints.md
```diff
@@ -0,0 +1,19 @@
+---
+layout: default
+title: JSON Endpoints
+---
+
+Lookbook currently implements a single JSON endpoint, reachable by adding the JSON format to the root Lookbook URL. It returns previews and their examples:
```
Spone, line 6: You ought to change the wording here to reflect the new route.

joelhawksley: Done!

allmarkedup:
Thanks for this @joelhawksley! I've considered adding an API/JSON endpoint along these lines before but never had a convincing use case for it. Sounds like there is one now :-)

I agree with @Spone, `/previews` makes more sense so this looks good to me now. I'm going to pull it down and give it a test quickly and then will get it out in a new release 😊

spec/requests/application_spec.rb
```diff
@@ -75,4 +75,18 @@
       end
     end
   end
+
+  context "JSON endpoint" do
```
Spone, line 79: Just a nitpick:


```suggestion
  describe "JSON endpoint" do
```

allmarkedup:
This is great, thanks guys, going to merge this now 👍

Spone (approved):

allmarkedup:
Just released this in v1.5.1 👍

PR #301, opened 2023-01-25
Stop sending `X-Frame-Options`

agrobbin:
Oftentimes, applications will customize [`X-Frame-Options`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) via `config.action_dispatch.default_headers`, setting it to `DENY` (similar to the `frame-ancestors` CSP directive). While a pretty good idea for most applications, Lookbook depends on rendering within an iframe, and so should not include that header!

This is similar to Lookbook not including `Content-Security-Policy`.

### DIFF
--- a/app/controllers/lookbook/previews_controller.rb
+++ b/app/controllers/lookbook/previews_controller.rb
@@ -3,6 +3,8 @@ module Lookbook
     include TargetableConcern
     include WithPreviewControllerConcern
 
+    before_action { response.headers.delete('X-Frame-Options') }
+
     layout "lookbook/inspector"
     helper Lookbook::PreviewHelper
 


allmarkedup:
Hey @agrobbin, many thanks for this - I hadn't considered this issue at all.

However I think it would probably be better to only include this in the controller that is responsible for rendering the preview itself ([`Lookbook::PreviewsController`](https://github.com/ViewComponent/lookbook/blob/main/app/controllers/lookbook/previews_controller.rb)), as that renders the only view that is displayed in a iframe. The other pages would then be left matching the host app's iframe policy.

Would you agree? If so it would be great if you could amend the PR to reflect this and then I'll get it merged in.

Thanks again for bringing this to my attention :-)

agrobbin:
@allmarkedup I've just updated this to only apply to `Lookbook::PreviewsController`!

allmarkedup:
@agrobbin looks great, will merge now - thanks for this!

agrobbin:
@allmarkedup you may already be aware of this, but it looks like this change is not included in the 2.0.0.beta.0 release.

allmarkedup:
@agrobbin Ah I must have missed that - I let the two branches diverge too much to rebase sanely so I've been cherry-picking commits from main to the v2 branch by hand for a month or so. Not the best system! 🙄

I'll make sure it gets pulled into the next beta release, thanks for the heads up.

agrobbin:
No worries! I was playing around locally, and it actually looks like the new `preview_embeds.policy` configuration is taking effect here, for reasons I'm not totally sure I follow. If it's intentional that `preview_embeds.policy` is used for rendering the preview itself (in `PreviewsController`), then this might Just Work, but if not, there might be a subtle bug with the preview embed functionality.

agrobbin:
I think it's because `PreviewsController` ends up calling `PreviewController` here:

* https://github.com/ViewComponent/lookbook/blob/v2.0.0-beta.0/app/controllers/lookbook/previews_controller.rb#L39
* https://github.com/ViewComponent/lookbook/blob/v2.0.0-beta.0/app/controllers/concerns/lookbook/with_preview_controller_concern.rb#L8
* https://github.com/ViewComponent/lookbook/blob/v2.0.0-beta.0/lib/lookbook/engine.rb#L165-L170
* https://github.com/ViewComponent/lookbook/blob/v2.0.0-beta.0/lib/lookbook/preview_controller_actions.rb

That last `PreviewControllerActions` concern executes [this `before_action`](https://github.com/ViewComponent/lookbook/blob/v2.0.0-beta.0/lib/lookbook/preview_controller_actions.rb#L8), applying the `X-Frame-Options` header.

I hope that's helpful!

allmarkedup:
Ah okay yes, I see the problem, that is very helpful thank you!

I think I'll open a new issue to remind me to pull in your change and also to get it playing nicely with the `preview_embeds.policy` option. Cheers @agrobbin :-)

PR #309, opened 2023-02-02
add preview_path to JSON endpoint

joelhawksley:
## Problem

I'd like to expose the preview_path for previews in the JSON endpoint I recently added.

## Proposed solution

Add a `preview_path` method to `Lookbook::Preview` and expose it via the API. I also exposed `lookup_path`. 

Note that in doing this I realized that the API does not handle preview groups, so I added some basic support for them here. @allmarkedup is infinite nesting allowed for preview groups? If so, we likely need to rethink this API entirely.

This controller method is getting a little messy. While I don't think we need to refactor it now, I think we're on the path towards needing a serializer or something similar.

### DIFF
--- a/app/controllers/lookbook/previews_controller.rb
+++ b/app/controllers/lookbook/previews_controller.rb
@@ -20,10 +20,15 @@ module Lookbook
               {
                 name: preview.name,
                 examples: preview.examples.map { |example|
-                  {
-                    inspect_path: example.url_path,
-                    name: example.name
-                  }
+                  case example
+                  when Lookbook::PreviewExample
+                    example_json(example)
+                  when Lookbook::PreviewGroup
+                    {
+                      name: example.name,
+                      examples: example.examples.map { |ex| example_json(ex) }
+                    }
+                  end
                 }
               }
             end
@@ -49,6 +54,16 @@ module Lookbook
       else
         show_404 layout: "lookbook/standalone"
       end
+    end
+
+    private
+
+    def example_json(example)
+      {
+        inspect_path: example.url_path,
+        name: example.name,
+        preview_path: example.preview_path
+      }
     end
   end
 end
--- a/app/controllers/lookbook/previews_controller.rb
+++ b/app/controllers/lookbook/previews_controller.rb
@@ -62,7 +62,8 @@ module Lookbook
       {
         inspect_path: example.url_path,
         name: example.name,
-        preview_path: example.preview_path
+        preview_path: example.preview_path,
+        lookup_path: example.lookup_path
       }
     end
   end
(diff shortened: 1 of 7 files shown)


app/controllers/lookbook/previews_controller.rb
```diff
@@ -20,10 +20,15 @@ def index
               {
                 name: preview.name,
```
jonrohan, line 21:
```suggestion
                name: preview.name,
                lookup_path: preview.lookup_path,
```

Wondering if we could add this small other thing too? Would end up adding `primer/alpha/component`

allmarkedup:
Nice one @joelhawksley, thanks for this - preview groups are not infinitely nestable so your code here will work fine.

I'm okay with merging this in like this for now but agree about cleaning it up a bit in the future.

As I mentioned before, I've explored some loose thoughts around this sort of thing before, although I was going down the route of exposing a more "full featured" dedicated JSON API at `/api` or so. I prototyped some code that implemented `toJSON` methods on each of the entities (called recursively to generated the JSON tree) and that seemed to work pretty well so that is what I'd probably pick up on again. But also very happy to hear any other thoughts on how best to implement something like this.

Right now I'm (still!) focussing on getting a beta of the work I've been doing on v2 out, once that is released I would definitely like revisit this with a bit more of a clear plan of what it would make sense to expose and how best to do it. So probably makes sense to save cleaning it up until then I think.

joelhawksley:
Sounds good @allmarkedup. I think this is a good place to leave this work for now, then. The tests are very high-level so we can refactor underneath them without much trouble.

lib/lookbook/entities/preview_example.rb
```diff
@@ -57,6 +57,10 @@ def url_path
       lookbook_inspect_path(path)
     end
 
+    def preview_path
```
allmarkedup, line 60: For consistency I think this should be added to the `PreviewExample` class as well as they are intended to be API-compatible where it makes sense to be.

I also think this method should probably accept a params hash that is passed to the path helper - obviously not useful for the JSON generation but then this method could be used in the views to replace calls like this: https://github.com/ViewComponent/lookbook/blob/f3cffecd9a317efba0e54bf5c10b27cba3fa7757/app/views/lookbook/inspector/show.html.erb#L54

Which again would make things more consistent.

Spone:
> any other thoughts on how best to implement something like this

I guess it depends if we want to rely on a dependency to render the JSON, such as [`jbuilder`](https://github.com/rails/jbuilder) or (my current favorite) [`jb`](https://github.com/amatsuda/jb).

allmarkedup:
Thanks @Spone I'll take a look at those 👍

lib/lookbook/entities/preview_example.rb
joelhawksley: @allmarkedup did you mean a different class? This is `PreviewExample`.

allmarkedup: Urgh sorry @joelhawksley - yes I meant `PreviewGroup`!

PR #311, opened 2023-02-04
Allow to nest previews in module with same name as parent component preview

phikes:
We use Lookbook in our project to document complex components. These components have child components nested in a module that's named after the complex parent component.

# Example

* `DashboardComponent`
* `Dashboard::UserListComponent`
* `Dashboard::ActivityComponent`

This allows us to keep our folder structure navigable and clean. 

# Problem

We would like to nest component previews as well.

## Example

Consider this example (with some of our domain data sprinkled in, but you get the point).

Screenshot 2023-02-04 at 11 28 20

While the navigation works perfectly fine, the preview renders a "404". **Note:: This works in `view_component` without Lookbook, which I think makes a strong point for supporting it here as well.**

# Solution

The problem was that the cache in a hierarchical collection considered a node as either having content OR having children, however in reality it can have both. With my changes this is supported. I also added tests (implicitly I test my case by adding a nested component preview which - without my changes - makes all tests for the nested components fail.).

If you need any more information or want to request changes, let me know. Thanks for the project! ❤️


allmarkedup:
Thanks for this @phikes! It definitely sounds like a use case Lookbook should support. I'll have a quick play with it and let you know if I think it needs anything else or if it can be merged in 😊

phikes:
Sure, lmk if I can help somehow. For the test failure we are seeing here: I saw that locally as well with a pristine repo :-(

allmarkedup:
Great thank you - not sure what is going on with that test failure but I'll try to get that fixed up now too :-)

allmarkedup:
Hey @phikes - so I've taken a look at this and had a bit more of a think about it.

The failure is because the change you made inadvertently causes the `Lookbook.previews` array to contain a mix of both previews *and* examples in some cases, which is not correct. In fact I think the fact it doesn't blow up in a lot more places probably says more about the sparseness of my test coverage rather than anything else ;-) Regardless, I think the change probably needs to happen in the place where the entity tree is constructed (`EntityTreeBuilder`). But I suspect changing anything in there will cause a lot of downstream issues.

I think I will need to take a little more time to think through how best to implement this in a way that doesn't cause potential issues elsewhere. It doesn't seem like a big change at first glance, but it breaks the current assumption that `previews` can only contain `preview examples` (at least in terms of the navigation) which definitely could cause some interesting side effects!

This is something I definitely want to support - I really like the idea of being able to organise previews in a way that helps to visually reinforce that a particular higher-order component has a number of subcomponents that are utilised by it. But it's getting late and I need a bit of a fresher brain than the one I currently have in to thing this through properly :-)

One thing - could you send me a screenshot of (the relevant portion of) the folder structure in your previews directory? I couldn't actually replicate the navigation tree in your screengrab above (although I may just be missing something obvious here!).

Thanks again for opening a discussion around this and once I have had a bit more of a chance to think it through I'll update you so we can get it moving forwards.

phikes:
Ahh, that makes sense. I did tinker with the source files in the gem directory before, so maybe that's why I saw the failure earlier. 

Thanks for the thorough explanation and investigation. I'll attach a screenshot. The nested component in the dummy folder of my fork should result in the same nav as well. Lmk if I can be of any further help.

Screenshot 2023-02-05 at 14 07 32

allmarkedup:
Thanks for the screenshot @phikes. However I'm afraid I've had a crazy week so haven't had a chance to think any more about this yet. I'll try to get to it as soon as I can.

allmarkedup:
@phikes I just wanted to let you know that I haven't forgotten about this! I'm just trying to focus on getting the v2.0 release out the door and after that is done I'll try and loop back round on this, as it's still definitely something I'm interested in supporting. Sorry it's taking so long to get to.

phikes:
Thanks so much for letting me know, I really appreciate your communication here :-)

allmarkedup:
@arg would you be interested in sharing your patch? I need to find some time to revisit this properly but it would be interesting to see how you have addressed it in the meantime.

agaman:
I have the same use-case and am using a similar fix as proposed.

```ruby
# lib/lookbook/entities/collections/concerns/hierarchical_collection.rb

def collect_ordered_entities(start_node)
  start_node.flat_map do |node|
    node_entity = node.content? ? [node.content] : []
    child_entities = collect_ordered_entities(node)
    [*node_entity, *child_entities]
end
```

allmarkedup:
@phikes @agaman @arg I finally found time to sit down today to take a look at this and I believe I've got a solution in the PR here: https://github.com/ViewComponent/lookbook/pull/476

It uses @agaman's amended `collect_ordered_entities` method (thanks! 😄) combined with some filtering to ensure that when calling `Lookbook.previews` only `PreviewEntity` objects are returned (without which all sorts of 'interesting' issues could appear down the line!).

Could you try pulling in the Lookjbook gem directly from the `nested-component-previews` branch and let me know if it fixes the issue for you (and doesn't cause any others)? I'm going to test it a little more throughly now with some other setups myself but I'd appreciate if you can give it a spin.

@phikes once again sorry for how long it has taken to get to this point! But hopefully this will mean we can finally close this down :-)

agaman:
The `nested-component-previews` branch is working for me. I haven't noticed any other issues, either. 👍 

Thanks for following up, @allmarkedup!

allmarkedup:
@agaman okay great, thanks for letting me know. All seems good from my point of view too so I'm going to go ahead and merge that PR in and close this one down. The next release will have the changes in it.

Thanks for the input everyone, much appreciated 😊

phikes:
Thanks for your work everyone! 🙌 I'll check next week at work if this works as intended for us.

allmarkedup:
Sorry for the delay in getting this released folks (moving house!) but it is now out in v2.0.5 - @phikes let me know how you get on if you manage to give it a try :-)

PR #317, opened 2023-02-07
Fix typo in README.md

dmarcoux:
https://www.dictionary.com/e/its-vs-its/


allmarkedup:
Thanks @dmarcoux! I'm absolutely terrible with it's vs its. No matter now many times I try to get it right I always miss it in so many places. Once someone opened a PR that corrected 20 individual instances of it in some docs I wrote 😂 So very much appreciate you taking the time to put it right here 👍

PR #319, opened 2023-02-12
Allow preview embeds to be disabled

agrobbin:
While I definitely understand the appeal of the new preview embeds functionality, it would be great if the feature could be disabled in case you don't want to expose that kind of functionality!

### DIFF
--- a/app/views/lookbook/inspector/show.html.erb
+++ b/app/views/lookbook/inspector/show.html.erb
@@ -33,13 +33,15 @@
 
         <% toolbar.with_section divide: :left, class: "flex-none relative z-10" do %>
           <%= lookbook_render :button_group do |group| %>
-            <%= group.with_button id: "embed-generator-dropdown-button", icon: :code_2, tooltip: "Get embed code" do |button| %>
-              <% button.with_dropdown.with_content(lookbook_render :embed_code_dropdown,
-                pages: @pages,
-                preview: @preview,
-                target: @target,
-                params: request.query_parameters
-              ) %>
+            <% if Lookbook.config.preview_embeds.enabled %>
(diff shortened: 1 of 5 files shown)


allmarkedup:
Thanks for this @agrobbin. I had initially considered adding config option for exactly this but felt like it was covered by the `preview_embeds.policy` as setting it to `DENY` effectively disables embeds.

However, looking at your PR I'm realising that just my approach would still leave the preview embed code dropdown in the inspector toolbar which is confusing. And only adding the embed routes if embeds are actually enabled makes sense too.

Even with the `preview_embeds.enabled` option you suggest I think those two issues should be addressed when the policy is set to `DENY`. Which feels like duplication... but I do also suspect that people would appreciate a simple, clear boolean toggle for disabling embeds as it's more obvious that using the `policy` setting.

So, what I think what I'm going to do is merge your PR, but I may come back to it and rework so that it disabling embeds in this way actually just sets the `policy` to DENY (under the hood). And then make the adding of routes and the display of the dropdown dependent on the value of policy option, not the enabled option. Does that make sense?

(The reason I want to keep the policy option and not just use a boolean toggle is that it allows for restricting use of embeds to the same origin, so that they can be used within documentation pages within Lookbook itself while still being disabled externally.)

How do you feel about embeds being enabled by default? Do you think that is something that should be opt-in rather than opt-out?

allmarkedup:
@agrobbin I think my brain is a little fried today - of course, as per the discussion above, if embeds are disabled entirely by default then they won't work in Lookbook documentation pages, which I don't feel is a desirable default (and would break existing behaviour). So I'm going to leave them as enabled by default, with the default policy set to "SAMEORIGIN", which covers all the bases I believe.

PR #333, opened 2023-02-16
Updating renders many calls to use with

leighhalliday:
v3 of ViewComponent requires `renders_many` slots to use `with_tab` rather than just `tab`. It looks like most (or all) of the other places in the codebase already used `with_` to call, but there were a few places within tabs that didn't.

### DIFF
--- a/app/views/lookbook/pages/show.html.erb
+++ b/app/views/lookbook/pages/show.html.erb
@@ -5,11 +5,11 @@
   @dom:update-complete.window="Lookbook.initEmbeds();">
   <div class="h-full bg-lookbook-page-bg relative">
     <% unless @error %>
-      
+
       <div class="absolute top-0 right-0 pt-1 pr-0 pl-1 pb-1 rounded-bl-md">
         <div class="bg-lookbook-page-bg opacity-90 absolute inset-0 w-full h-full z-0"></div>
         <div class="relative z-10 flex items-center">
-        
+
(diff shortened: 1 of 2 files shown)


app/views/lookbook/pages/show.html.erb
```diff
@@ -5,11 +5,11 @@
   @dom:update-complete.window="Lookbook.initEmbeds();">
   <div class="h-full bg-lookbook-page-bg relative">
     <% unless @error %>
-
```
Spone, line 8: The whitespace changes in this file seem unintentional

Spone (commented):
Thanks for spotting this! Just a quick comment.

app/views/lookbook/pages/show.html.erb
leighhalliday: Gah, my editor strips whitespace on save :( It wasn't intentional, but I don't think it would cause a problem either.

allmarkedup:
Thanks @leighhalliday! I really thought I'd got all of them so I'm glad your eyes are sharper than mine. Much appreciated :-)

app/views/lookbook/pages/show.html.erb
allmarkedup: No worries, I'll merge like this and it should get automatically removed again next time I do a release (or before!).

PR #342, opened 2023-02-20
docs: Allow method_list titles to be used as anchors

Spone:
Currently titles in `toc` are clickable, but `method_list` titles do not behave as anchors (they don't have an ID), so nothing happens when you click them.

image

image


allmarkedup:
thanks @Spone! Good catch :-)

PR #363, opened 2023-03-17
Use double quotes to fix escape issues

fusion2004:
I ran into an issue where trying to put a paragraph of text including single quotes (`'`) as the default value for a param ended up with the text param showing "undefined" in the UI. Looking into `json_escape`, it is expecting to escape double quotes not single quotes, so I changed the code here to use escaped double quotes instead of single quotes.

Running this locally fixes the issue for me.


allmarkedup:
Hey @fusion2004 this is great, thanks for this! I'll merge this in now. There is supposed to be a test for characters in params but it's obviously missing something so I'll update it to cover this. 

Thanks again :-)

PR #371, opened 2023-03-20
Preview Output Transformation

fusion2004:
So this is my take on an idea to enable full manipulation of the outputted HTML for a given Preview...

The reason I'm looking for this capability is that I am trying to build a component system for emails using Phlex & Lookbook, and I want the output to more closely match the processing we do on outgoing emails to inline CSS and automatically set some attributes on HTML elements.

With this change, I can subclass all my email components from a base class where I override `render(...)` to add the default `:output_transformation` they'll all use, and when looking at any individual preview you won't see any of that in the Source panel, but the transformation will be reflected in the Preview & HTML panels.

I was first trying to implement this through the `layout`, which technically worked, but the inlined CSS would only be visible on the Preview panel and not the HTML panel.

Do you think this is a good approach to implement this feature, or do you have suggestions of a different way to implement this?

### DIFF
--- a/app/views/lookbook/previews/preview.html.erb
+++ b/app/views/lookbook/previews/preview.html.erb
@@ -1,5 +1,11 @@
-<% if @render_args[:component] %>
-  <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
-<% else %>
-  <%= render(@render_args[:template], **@render_args[:locals], &@render_args[:block]) %>
-<% end %>
+<%=
+  output = if @render_args[:component]
+    render(@render_args[:component], @render_args[:args], &@render_args[:block])
+  else
+    render(@render_args[:template], **@render_args[:locals], &@render_args[:block])
+  end
+
+  output = @render_args[:output_transformer].call(output) if @render_args[:output_transformer]
+
+  output
+%>
--- a/spec/dummy/test/components/previews/partial_example_preview.rb
+++ b/spec/dummy/test/components/previews/partial_example_preview.rb
@@ -6,4 +6,10 @@ class PartialExamplePreview < Lookbook::Preview
   def helpers
     render "partials/helpers_example"
   end
+
+  def transform
+    render_args = render "partials/basic_example"
+    render_args[:output_transformer] = ->(html) { "<span>transformed</span>#{html}".html_safe }
+    render_args
+  end
 end
--- a/spec/dummy/test/components/previews/phlex_example_preview.rb
+++ b/spec/dummy/test/components/previews/phlex_example_preview.rb
@@ -6,4 +6,10 @@ class PhlexExamplePreview < Lookbook::Preview
   def helpers
     render Views::Phlex::HelpersExample.new
   end
+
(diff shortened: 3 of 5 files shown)


fusion2004:
The CI failures are interesting to me... I wasn't even aware that lookbook worked with rails 6.0! What is different about how it renders the partials when on rails 6.0?

allmarkedup:
Hey @fusion2004, this sounds like a really interesting project and use-case!

So I've actually prototyped something along these lines before myself - but the approach I was experimenting with was implementing post-render manipulation via a `Lookbook.after_render` [lifecycle callback](https://v2.lookbook.build/extend/other/hooks) that receives the result from the rendering of the `app/views/lookbook/previews/preview.html.erb` template and returns the transformed HTML. Something like:

```rb
Lookbook.after_render do |html|
  run_transformation(html)
end
```

I was also tinkering with adding a `@after_render` tag that could be applied to a preview class or an individual scenario to specify a method for doing the transformation (in the same way as a method can be specified as the [provider of @param options](https://v2.lookbook.build/guide/previews/params#using-a-method). Something like:

```rb
# @after_render :my_transform_method
class ExamplePreview < Lookbook::Preview
  
  # preview scenarios here...  

  private

  def my_transform_method(html)
    run_transformation(html)
  end
end
```

The reason I never actually implemented any of these is that I was struggling to come up with a compelling use-case for it - one that you have clearly found :-)

Regarding your implementation, it is certainly something that you could do right now by adding the amended preview template within your app at `app/views/lookbook/previews/preview.html.erb` and then subclassing the component preview classes to add in the transformer to the `render_args` as you describe. Whilst overriding templates within an engine is never _completely_ safe from unexpected breaking changes I don't envisage updating that template unless I absolutely have to (in case others have overridden it in the same way!).

As a way to implement this feature more broadly (i.e. in Lookbook core), it definitely has it's merits (using a different transform for different previews could be achieved using different subclasses with their own transform methods, which would be nice) but I have to say my inclination would still be towards an approach that felt more like configuration and didn't require subclassing the `Lookbook::Preview` class to use it - something like the lifecycle callback or the preview tag approaches I mentioned above.

How would you feel about something along the lines of an `after_render` callback or a `@after_render` tag instead of this preview subclass approach? I didn't get very far with thinking through the edge cases so they may well not be the best approach but I'd be interested to know what you think. Is there anything there that would hinder you in implementing the functionality you are trying to build here?

I'm going to have a bit more of a think about your approach and try to compare it with my previous ideas around this. I do like how little code it involves!

And If I have a chance I'll try to dig out the stuff I prototyped and see if it's in a state where I can push up a branch for you to play with and see how it feels.

Thanks again for opening the PR - it's definitely a use case I'd like to cater for once we figure out the best way to do so :-)

fusion2004:
I think I really like the idea of the `@after_render` tag. I'll look into trying to implement that! And yea, any previous WIPs you've got to help me figure it out would be great, haha.

fusion2004:
New approach submitted as a separate PR #373!

PR #373, opened 2023-03-21
@after_render scenario output transform

fusion2004:
Based on the discussion in #371, here's another take on providing a way for folks to transform the HTML output of a scenario after running the render.

This time through a new tag: `@after_render`. It can be put on the preview class or the scenario method, with the method taking higher priority. You provide the name of a private method to call that will receive one argument, a string of the rendered HTML.

Example:
```ruby
class MyComponentPreview < Lookbook::Preview
  # @after_render :prepend_test
  def default
    render MyComponent.new
  end

  private

  def prepend_test(html)
    "<p>test</p>#{html}"
  end
end

### DIFF
--- a/lib/lookbook/tags/after_render_tag.rb
+++ b/lib/lookbook/tags/after_render_tag.rb
@@ -1,0 +1,7 @@
+module Lookbook
+  class AfterRenderTag < YardTag
+    def value
+      text.sub(/\A:/, "")
+    end
+  end
+end
--- a/app/controllers/concerns/lookbook/targetable_concern.rb
+++ b/app/controllers/concerns/lookbook/targetable_concern.rb
@@ -96,7 +96,7 @@ module Lookbook
 
       scenarios = @target ? @target.scenarios : []
       rendered_scenarios = scenarios.map do |scenario|
-        output = preview_controller.process(:render_scenario_to_string, @preview, scenario.name)
+        output = preview_controller.process(:render_scenario_to_string, @preview, scenario)
         RenderedScenarioEntity.new(scenario, output, preview_controller.params)
       end
 
--- a/config/tags.yml
+++ b/config/tags.yml
@@ -38,3 +38,7 @@ shared:
   source:
     label: Source file
     opts: {}
+
+  after_render:
+    label: After render
+    opts: {}
--- a/lib/lookbook/engine.rb
+++ b/lib/lookbook/engine.rb
@@ -45,6 +45,8 @@ module Lookbook
         if vc_config.view_component_path.present?
           opts.component_paths << vc_config.view_component_path
         end
+
+        ViewComponent::Preview.extend(Lookbook::PreviewAfterRender)
       end
 
       opts.reload_on_change = host_config.reload_classes_only_on_change if opts.reload_on_change.nil?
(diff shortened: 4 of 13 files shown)

### CODE REVIEW

**Comment by allmarkedup** [comment]
@fusion2004 woah you are fast! And this looks fantastic 😀

Just one thing that would be good to change - I think the method name should be specified with a colon prefix (symbol-style) so that it matches with the existing syntax for referencing a method when generating `@param` tag options: https://v2.lookbook.build/guide/previews/params#using-a-method

So `@after_render :prepend_test` instead of `@after_render prepend_test`. Does that sound okay?

I'm going to give this a spin now and I'll let you know if I have any more comments shortly. Really appreciate your contribution here!

**Comment by allmarkedup** [comment]
@fusion2004 Okay this is all working great. I like this a lot :-)

One small issue - this will currently only work for previews that extend `Lookbook::Preview`. A lot of Lookbook users are using 'regular' ViewComponent previews that extend `ViewComponent::Preview`, so it will need to be monkey patched to add in the `after_render` class method.

I don't have strong opinions on how this is done - happy to go with however you want to do it but it should probably be applied within this conditional here: https://github.com/ViewComponent/lookbook/blob/d2b4f83398e3139d453d968a9c03b45b4f3c183c/lib/lookbook/engine.rb#L36-L48

Obviously if there is a nice way to do it that doesn't introduce duplication with the same method in the `Lookbook::Preview` class then that would be great. But I'm not super worried about it as it's only a couple of lines.

I hope it wasn't too hard work trying to figure out how everything fits together in there!

**Comment by allmarkedup** [comment]
@fusion2004 I'm also happy to add the monkey patch in if you prefer - just let me know 👍

**Comment by fusion2004** [comment]
Ah cool, yea I'll make those 2 changes now.

File: lib/lookbook/tags/after_render_tag.rb [hunk 1]
```diff
@@ -0,0 +1,7 @@
+module Lookbook
+  class AfterRenderTag < YardTag
+    def value
+      text.sub(/\A:/, "")
```
fusion2004, line 4: Originally, I tried going down a path that utilized `Lookbook::MethodResolver` and similar and kept running into errors and infinite loops, so I gave up and went for a simple string substitution. :P

fusion2004:
@allmarkedup This is ready for another look!

lib/lookbook/tags/after_render_tag.rb
allmarkedup: Ha yeah I think the simple approach looks just fine here 🙂

allmarkedup:
@fusion2004 The changes all look good and I've just had another play and it's all working very nicely. Great stuff! I really appreciate the contribution 🙏

I'll merge this in and then add a section about this to the documentation. Unless you'd prefer to write it yourself? It's your feature so whatever you prefer, just let me know. Either way I'll merge this now and then the docs can be handled in a separate PR.

Thanks again for your time on this and for your thoroughness in implementation and testing. I like this feature a lot! 😄

fusion2004:
I wasn't sure on the best place in the docs to put it (and maybe the answer is it should go in multiple places), so if you want to tackle that it'd be much appreciated!

allmarkedup:
Of course, no worries at all - I'll try and get to that tomorrow, and this will go out in the next release, probably towards the end of the week.

I'll let you know when the docs are ready so you can take a look at them before they go up.

allmarkedup:
Hey @fusion2004 I just wanted to let you know that I released beta.9 last night which includes this feature. I've added some docs here: https://v2.lookbook.build/guide/previews/output - thanks again!

Issue #400, opened 2023-04-05
Support for Phlex builder-style components


willcosgrove:
### Describe the bug

Phlex views that use a builder-style API are rendering in unbuffered mode, as though they were called from an ERB template.

### To Reproduce

Given a pair of Phlex components defined like this:

```ruby
class List < Phlex::HTML
  def template
    ul { yield }
  end

  def item(...)
    render Item.new(...)
  end
end

class Item < Phlex::HTML
  def template
    li { yield }
  end
end
```

And a preview defined like this:

```ruby
class ListPreview < Lookbook::Preview
  def default
    render List.new do |list|
      list.item { "Hello" }
      list.item { "World" }
    end
  end
end
```

You will get an output that looks like this:

```html
<ul>
  <li>World</li>
</ul>
```

This is because when Rails' `render` method is used (which gets turned into `render_in`), Phlex will [turn the component into](https://github.com/phlex-ruby/phlex-rails/blob/7ee772c05d5b3133c9068ada0a618026f7e02e8d/lib/phlex/rails/sgml/overrides.rb#L38-L40) an ["unbuffered"](https://github.com/phlex-ruby/phlex/blob/5e0ca1f5faaf4ba2a78c61fdc26b166e1681d72d/lib/phlex/unbuffered.rb#L4) representation. This means that now each method that normally would have added output to the internal buffer and returned nil, will now _not_ append anything to the internal buffer, and instead return the string it would have appended.

So in our above example the line by line evaluation looks like this:

```ruby
render List.new do |list|
  list.item { "Hello" } # => "<li>Hello</li>
  list.item { "World" } # => "<li>World</li>
end
```

When the `List` component finishes it's block, it sees that it's own internal buffer has not been modified and assumes that the return value of the block is supposed to be appended to the buffer, which is why we get the final list item only.

This is how Phlex must work when it's being rendered from inside another template

```erb
<%= render List.new do |list| %>
  <li>First</li>
  <%= list.item { "Hello" } %>
  <%= list.item { "World" } %>
  <li>Last</li>
<% end %>
```

In this context, the buffer is being managed by ERB.

### Expected behavior

My expectation was that, because I am rendering my Phlex component in a (for lack of a better term) "ruby context"— not inside of a template, that the component would behave the same way it does when I render it within other Phlex views.

[...]

PR #404, opened 2023-04-06
Fix rendering of builder-style Phlex components

allmarkedup:
See discussion in issue #400.


willcosgrove:
Just tested this out on my project and it looks good to me! Thanks for your help with getting this working so quickly! 🙏

joeldrapper (commented):
This looks good to me. It's a shame we need to have a special condition here. I’m really open to putting a fix in `phlex-rails` itself, but I can’t think of a way to do it.

allmarkedup:
@willcosgrove great, thanks for testing it out and letting me know. I'll get this merged in now and out in the next release.

@joeldrapper thanks for taking a look - it is a bit of a shame that it needs special casing here (ideally like to keep this template as simple as possible as it's conceivable that people might choose to override it in their app) but it's not a big deal really. I guess more generally for me it feels a little unintuitive that a component instantiated with a block in a 'Ruby' context can't then be rendered in a template... but I realise that is probably not a very common use case for most people anyway.

PR #412, opened 2023-04-10
DO NOT MERGE - Use experimental CSP-friendly Alpine package

joelhawksley:
We’re heavy users of Lookbook at GitHub. In addition to our public Primer ViewComponents Lookbook (https://primer.style/view-components/lookbook/inspect/primer/beta/blankslate/default/), we use Lookbook internally for our application-specific components. Recently, we’ve been looking into deploying our internal Lookbook to production.

One blocker to doing so is resolving the compromised CSP configuration we've had to use to run Lookbook, which requires unsafe-eval due to usage of AlpineJS. In looking into this issue, I found that Alpine has an experimental CSP-friendly package. I've added it in this PR for the sake of testing by others to see if anything is broken.

How does this look to you @allmarkedup?

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -1,4 +1,4 @@
-import Alpine from "alpinejs";
+import Alpine from "@alpinejs/csp";
 
 import Morph from "@alpinejs/morph";
 import Persist from "@alpinejs/persist";
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,11 +5,11 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "lookbook",
       "version": "2.0.0-rc.2",
       "license": "MIT",
       "dependencies": {
         "@alpinejs/collapse": "^3.10.5",
+        "@alpinejs/csp": "https://gitpkg.now.sh/alpinejs/alpine/packages/csp?main",
         "@alpinejs/morph": "^3.10.5",
         "@alpinejs/persist": "^3.10.5",
         "@rails/actioncable": "^6.1.4",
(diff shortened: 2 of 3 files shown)


allmarkedup:
Thanks for this @joelhawksley - although I've been away for a few days around Easter and am just catching up with everything so I haven't had a chance to look at this in detail yet I'm afraid!

The failing tests here are not to do with your PR, they are to do with a change I pushed just before I went away that has broken the main v2 branch. I'll hopefully get that fixed up in the next day or two.

With regards to the CSP-friendly build specifically, it's something that has been mentioned before, unfortunately it seems like it hasn't been updated in a while and there is no clear roadmap to it having an official release. Having said that, I wouldn't have any problem with building it from source and vendoring it in in it's current state if we can get it working (the `gitpkg` approach here is perfect for testing but I'd prefer to not rely on it for if we ended up using it in a proper release).

I'm pretty sure that the restrictions around its use will currently cause issues in a few places, but off the top of my head I can't see why it wouldn't be possible to work around those. I'll need to take a little time to refresh my memory of what exactly that would entail - I'll try and do that over the next few days alongside giving it this PR a spin.

Thanks for bringing this up again, I know it's something that others would also appreciate getting working so I'll try and prioritise at least a first exploration to see how much work it would be. Depending on what shakes out of that I may try and squeeze it in to the v2 release, but I also don't want to hold that up too much at this late stage so it may have to be handled in a later follow-up release if the changes needed are significant. I'll let you know how I get on :-)

allmarkedup:
Just a followup note here - the assets need rebuilding after this change, and that rebuild is currently failing (not reflected in the tests here because currently the build needs to be run manually and doesn't happen as part of a GH actions workflow - something that is on my list of things to improve!).

To run the asset build `npm run build` needs to be run after the packages are updated.

I haven't had a chance to dive in properly yet but It looks like the `gitpkg` build step is not working correctly for the CSP directory? I'll take a look over the weekend and see if I can figure it out.

allmarkedup:
@joelhawksley after some investigation it seems that updating the existing UI to use the more restrictive CSP build is not really feasible. It needs reworking from the ground up with support for that built in from the start.

It _is_ still something I'm keen to support, but it will likely be as part of a bigger UI rework. I've actually started exploring what this might look like already but I don't have anything ready to show just yet. 

So I'm going to go ahead and close this down for now but it will stay firmly on my radar :-)

PR #418, opened 2023-04-19
Bug Fix: Error message title not wrapping causes text overflow

NYDrewReynolds:
Ran into an error page and noticed the error message ran off the side of the page. This change removes `white-space: nowrap;` from `[data-component="message"] .message-title` to allow for reading the entire error.

**BEFORE**
Screenshot 2023-04-19 at 1 25 11 PM

**AFTER**
Screenshot 2023-04-19 at 1 25 27 PM

### DIFF
--- a/public/lookbook-assets/css/lookbook.css
+++ b/public/lookbook-assets/css/lookbook.css
@@ -1526,7 +1526,6 @@ pre[class*="language-"] {
 
 [data-component="message"] .message-title {
   text-overflow: ellipsis;
-  white-space: nowrap;
   text-transform: uppercase;
   letter-spacing: .025em;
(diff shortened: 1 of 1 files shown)


allmarkedup:
@NYDrewReynolds many thanks! Appreciate you taking the time out to fix this and open a PR 👍

allmarkedup:
@NYDrewReynolds Ah I've actually just noticed that you've fixed this in the compiled CSS but not in the source CSS for that component, so it will get overwritten the next time that the assets are built. No worries though - I'll merge this anyway so you get the contribution credit and then I will make a separate followup commit to change it in the message component CSS source file. Thanks again :-)

PR #422, opened 2023-04-22
Fix multi-line parameter defaults

GabeIsman:
I noticed a bug with the new dynamic parameter handling feature where default values containing newlines were breaking the JS. This PR addresses that very simply by escaping newlines. I don't know enough about AlpineJS to know if there's a more elegant way to accomplish this escaping. Happy to adjust if there's something cleaner.


allmarkedup:
Hey @GabeIsman many thanks for this. I think this looks good as a fix for now, the whole parameter escape process definitely needs a clean up and extracting into a service or similar where it can be better tested but until then this is just fine.

Appreciate you taking time to figure the issue out and open the PR, thanks again 👍

PR #446, opened 2023-04-30
Fix two small bugs in param tag options parsing.

GabeIsman:
This PR fixes a two small bugs that I noticed in how descriptions are handled in `@param` tags, particularly when there are tag options containing double-quoted strings.

1. The `DESCRIPTION_MATCHER` regex in `ParamTag` was greedy-matching the paired `"` character. It should be non-greedy. I was getting a description of `Whether this should be on or not" { value_type: "Boolean` on my parameter in the UI.
2. The `DESCRIPTION_MATCHER` was getting passed to `gsub!` rather than `sub!` when it matched, so the tag option string getting passed to `YardTag` looked like `{ choices: [primary, secondary], value_type: }` instead of `{ choices: [primary, secondary], value_type: "Symbol" }` because the `"Symbol"` was being removed by `gsub`.

I added a couple of tests for options in the `ParamTag` spec. They will both fail if the implementation changes in here are removed. I wouldn't be surprised at all if there were some other odd bugs that could be turned up with some more thorough tests though!

### DIFF
--- a/lib/lookbook/tags/param_tag.rb
+++ b/lib/lookbook/tags/param_tag.rb
@@ -1,7 +1,7 @@
 module Lookbook
   class ParamTag < YardTag
     VALUE_TYPE_MATCHER = /^(\[\s?([A-Z]{1}\w+)\s?\])/
-    DESCRIPTION_MATCHER = /(?<=\s|^)"(.*[^\\])"(?:\s|$)/
+    DESCRIPTION_MATCHER = /(?<=\s|^)"(.*?[^\\])"(?:\s|$)/
 
     supports_options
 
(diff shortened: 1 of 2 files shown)


allmarkedup:
@GabeIsman many thanks for this! My regex-fu is pretty poor so PRs like this are much appreciated :-) There is probably a much better way to do the `@param` tag string parsing tbh but it might be a while before I get around to reworking it.

I'll check it out in a bit more detail now and then get it merged and out in the next release. Thanks again 🙏

allmarkedup:
@GabeIsman this all looks good... although diving back into that parsing code makes me want to cry a bit 😂

I've found another bug while testing out your fixes (**not** as a result of your fixes though!) but I'm going to merge this and then fix it in a separate PR/commit.

Thanks :-)

GabeIsman:
Thanks for reviewing and merging @allmarkedup! Happy to help out as I can.

PR #466, opened 2023-05-16
Fix broken link

dkniffin:


dkniffin:
Hmm, that change didn't seem to fix it, according to the deployed preview. Here's where that link shows up:

Screen Shot 2023-05-16 at 12 12 33 PM

allmarkedup:
Thanks @dkniffin - it should actually be `<%= extend_url :params_overview %>`.

Embarrassingly I have actually just accidentally just included a fix for this in PR #467 (which hopefully addresses your Nokogiri issue!) so I'll just close this for now and when that PR gets merged it will fix the link too :-)

Thanks for bringing it to my attention 👍

PR #471, opened 2023-05-23
Silence the `FileWatcher` from always logging the type of watcher

agrobbin:
@allmarkedup I hope you're OK with this proposed change! On application startup, you see one of these messages twice, and it seemed a little chatty.

If someone wants to know which watcher that is used, they can look at `Lookbook.debug_data[:dependencies][:lookbook]` or `Lookbook::Engine.auto_refresh?`.


allmarkedup:
I'm fine with this @agrobbin! Good call, thank you 👍

PR #487, opened 2023-06-12
Prefer Enumerable#grep to filter by type

olleolleolle:
This reformulates a tiny part of the code to use an existing Ruby method.


allmarkedup:
Hey @olleolleolle, really sorry it's taken me so long to take a look at this.

I did not know about `Enumerable#grep` but this is so much nicer, thank you! The only problem is now I want to go through all my Ruby code to see where I can use it to clean things up a bit... :)

Will merge now, much appreciated.

PR #503, opened 2023-07-31
Disable code reloading in production by default

natematykiewicz:
By default a Rails app configs have `cache_classes` as false in develoment and true in production, and does not change `reload_classes_only_on_change` from its default `true`.

This means that Lookbook currently actually watches for changes in Production as well as Development, because it's not taking `cache_classes` into account.

To quote the [Rails Guides](https://guides.rubyonrails.org/threading_and_code_execution.html#configuration):
> The Reloader only checks for file changes when `cache_classes` is false and `reload_classes_only_on_change` is true (which is the default in the `development` environment).
>
> When `cache_classes` is true (in `production`, by default), the Reloader is only a pass-through to the Executor.

I discovered this when my app wouldn't boot in production and instead had this error:
```
[1] ! Unable to load application: Errno::EMFILE: Too many open files - Failed to initialize inotify: the user limit on the total number of inotify instances has been reached.
bundler: failed to load command: puma (/app/vendor/bundle/ruby/3.2.0/bin/puma)
/app/vendor/bundle/ruby/3.2.0/gems/rb-inotify-0.10.1/lib/rb-inotify/notifier.rb:69:in `initialize': Too many open files - Failed to initialize inotify: the user limit on the total number of inotify instances has been reached. (Errno::EMFILE)
    from /app/vendor/bundle/ruby/3.2.0/gems/listen-3.8.0/lib/listen/adapter/linux.rb:29:in `new'
    from /app/vendor/bundle/ruby/3.2.0/gems/listen-3.8.0/lib/listen/adapter/linux.rb:29:in `_configure'
    from /app/vendor/bundle/ruby/3.2.0/gems/listen-3.8.0/lib/listen/adapter/base.rb:47:in `block in configure'
    from /app/vendor/bundle/ruby/3.2.0/gems/listen-3.8.0/lib/listen/adapter/base.rb:42:in `each'
    from /app/vendor/bundle/ruby/3.2.0/gems/listen-3.8.0/lib/listen/adapter/base.rb:42:in `configure'
    from /app/vendor/bundle/ruby/3.2.0/gems/listen-3.8.0/lib/listen/adapter/base.rb:66:in `start'
    from /usr/lib/fullstaq-ruby/versions/3.2.2-jemalloc/lib/ruby/3.2.0/forwardable.rb:240:in `start'
    from /app/vendor/bundle/ruby/3.2.0/gems/listen-3.8.0/lib/listen/listener.rb:71:in `block in <class:Listener>'
[...]

PR #514, opened 2023-08-03
Disable reload_classes_only_on_change if cache_classes is enabled

stanhu:
Rails will ignore `config.reload_classes_only_on_change` if `config.cache_classes` is
enabled. https://guides.rubyonrails.org/configuring.html#config-reload-classes-only-on-change: states:

>>>
3.2.30 config.reload_classes_only_on_change

Enables or disables reloading of classes only when tracked files change. By default tracks everything on autoload paths and is set to true. If config.cache_classes is true, this option is ignored.
>>>

It was surprising to us that even though `cache_classes` were set to true that lookbook still instantiated file watchers.

### DIFF
--- a/docs/src/_data/config_options.yml
+++ b/docs/src/_data/config_options.yml
@@ -223,7 +223,7 @@ debugging:
 system:
   - name: reload_on_change
     types: Boolean
-    default: "config.reload_classes_only_on_change"
+    default: "!config.cache_classes && config.reload_classes_only_on_change"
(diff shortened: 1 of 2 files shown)


allmarkedup:
Hey @stanhu, apologies for the slow reply and many thanks for this. This was a mistake on my part which has now been fixed in #503 and will be included in the next release.

I'll close this down now but thanks again for taking the time to open the PR, much appreciated.

PR #515, opened 2023-08-25
Fix attribute escaping deprecation warning

stefsava:
escape_attribute_key has also been deprecated in the 5.2 branch since version 5.2.7.1


allmarkedup:
@stefsava good to know! Thanks for the PR, much appreciated :)

PR #516, opened 2023-08-25
Fix accessibility issues

tysongach:
The Lookbook UI had a few accessibility issues flagged by [axe], an accessibility testing tool:

Problem: Buttons must have discernible text
Element: Button that opens the debug menu in the app header 
Ref: https://dequeuniversity.com/rules/axe/4.6/button-name 
Solution: Provide a tooltip, which is used as the button’s accessible name

Problem: IDs of active elements must be unique
Element: The two ‘show drawer’ buttons
Ref: https://dequeuniversity.com/rules/axe/4.6/duplicate-id-active
Solution: Give each button a unique ID

Problem: Zooming and scaling must not be disabled
Ref: https://dequeuniversity.com/rules/axe/4.6/meta-viewport 
Solution: Remove `maximum-scale=1.0` and `user-scalable=0` from the `viewport` metadata

[axe]: https://www.deque.com/axe


allmarkedup:
Hey @tysongach, many thanks for this and so sorry for the embarassingly slow reply. I know Lookbook's accessibility is pretty poor in places so I really appreciate PRs like this one.

I'll merge it in now and the changes will be included in the next release. Thanks again!

PR #517, opened 2023-08-27
Avoid problem with Rails classic autoloader

PauloMiranda98:
For some reason, in our project, we have a problem in the `classic` autoloader (we haven't migrated to `zeitwerk` yet). This problem caused the lookbook to use our `ApplicationController` instead of using `lookbook::ApplicationController`, generating an error to render the components. With that, to avoid problems with other people, we made this correction.


allmarkedup:
@PauloMiranda98 sorry for being so slow to reply and many thanks for this, looks great. Will merge it in now and get it out in the next release, hopefully in the next couple of days.

Issue #528, opened 2023-09-18
Support better errors / web console instead of rescuing errors in Lookbook


wheeyls:
I use better errors in development, and I prefer the experience to the customized error pages that Lookbook provides.

I don't see an option to remove Lookbook's error handling and let me take care of it. Am I missing this, or can it be added?

allmarkedup:
Hey @wheeyls - no I'm afraid this is not possible right now, sorry 😕 I'm aware Lookbook's error handling is not always the best so this would be a really nice feature.

It would be nice if Lookbook could detect if it was being used in the parent app and somehow automatically defer to it for error handing... or even bundle it as the default error handler, perhaps.

I'll leave this issue open as a feature request and try to have a bit of a think about it when I have some time. Thanks for the suggestion!

allmarkedup:
Config option added in #582 (thanks @wheeyls!) has been merged in and will go out in the next release so going to close this down now.

PR #582, opened 2024-01-28
Allow disabling of error handling

wheeyls:
Per issue #528
Allow opting out of catching errors in the application controller. Allows usage of web_console, better_errors, and other solutions.

### DIFF
--- a/app/controllers/lookbook/application_controller.rb
+++ b/app/controllers/lookbook/application_controller.rb
@@ -57,6 +57,8 @@ module Lookbook
     end
 
     def handle_error(err)
+      raise err if Lookbook.config.preview_disable_error_handling
+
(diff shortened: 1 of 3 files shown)


allmarkedup:
Hey @wheeyls, thank you for this!

In a future major release of Lookbook it may be that not handling errors becomes the default behaviour as I've never been super happy with it. But having it as a config option for now is great.

I'll just give it a try now but if I don't find any issues with it I'll merge it in, add a note in the docs and get it out in the next release. Thanks again :)

allmarkedup:
Looks good, merging now - thanks again for the contribution!

wheeyls:
Thanks @allmarkedup.And yeah I agree as a rule I'd prefer the default behavior is to let me choose my error handling strategy myself, there are lots of other solutions out there.

Issue #584, opened 2024-01-31
Better integration with Phlex


stephannv:
I'm facing a issue with Lookbook + Phlex:

Phlex allows you to write something like this:
```ruby
render Views::Phlex::ListExample.new do |list|
  list.item { "Hello" }  
  render(Views::Phlex::Item.new) { "Help!"}
  list.item { "World" }
end
```

But when I write this example on a Lookbook::Preview class, it doesn't work. To reproduce the error, paste the code above into `builder` scenario on this file: https://github.com/ViewComponent/lookbook/blob/main/spec/dummy/test/components/previews/phlex_example_preview.rb#L10.

The `render(Views::Phlex::Item.new) { "Help!" }` returns a hash because it is using `render` from Lookbook::Preview and the component isn't rendered. To solve that I could write something like this: 
```ruby
render Views::Phlex::ListExample.new do |list|
  list.item { "Hello" }  
  list.render(Views::Phlex::Item.new) { "Help!"}
  list.item { "World" }
end
```

Then I will have two problems:
- On source tab, will display component usage `list.render ...`, but it isn't required, so other devs could understand that unnecessary code is required to build the component
- I cannot use my component helpers. I added helpers to Lookbook::Preview, eg.:
```ruby
module DesignSystemHelpers
  def my_button(*args, **kwargs, &block)
    render DesignSystem::ButtonComponent.new(*args, **kwargs, &block)
  end
  
  def my_card(...)
    ...
  end
  
  ...
end

# on initializers
config.to_prepare do
  Lookbook::Preview.include DesignSystemHelpers
end
```
So something like this doesn't work on Lookbook:
```ruby
def default
  my_card do |card| # this will be rendered
    card.actions do # this will be rendered
      my_button(...) # this will NOT be rendered because Lookbook::Preview `#render`
      my_button(...) # this will NOT be rendered because Lookbook::Preview `#render`
    end
  end
end
```
This is important because I think Lookbook is a source of how developers should use our components.

And the other problem is that I cannot do that on preview files:
```ruby
render Views::Phlex::ListExample.new do |list|
  h1 { "My list" }
  list.item { "Hello" }  
  list.item { "World" }
end
```

It will raise `undefined method 'h1' for an instance of MyComponentPreview`.

_Originally posted by @stephannv in https://github.com/ViewComponent/lookbook/issues/400#issuecomment-1918097069_

* Lookbook:  2.2.0
* Phlex-Rails: 1.1.1
[...]

PR #585, opened 2024-01-31
Support Phlex DSL on scenarios

stephannv:
Fixes: https://github.com/ViewComponent/lookbook/issues/584

I'm not sure if this is the best solution but it's working 😅

Instead evaluating the block on Lookbook::Preview context, it is evaluating on component context, so it can render other components and use Phlex DSL.

### DIFF
--- a/app/views/lookbook/previews/preview.html.erb
+++ b/app/views/lookbook/previews/preview.html.erb
@@ -1,9 +1,5 @@
 <% if @render_args[:component] %>
-  <% if defined?(Phlex::SGML) && @render_args[:component].is_a?(Phlex::SGML) %>
-    <%= raw(@render_args[:component].call(view_context: self, &@render_args[:block])) %>
-  <% else %>
-    <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
-  <% end %>
+  <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
 <% else %>
   <%= render(@render_args[:template], **@render_args[:locals], &@render_args[:block]) %>
 <% end %>
--- a/lib/lookbook/phlex_preview.rb
+++ b/lib/lookbook/phlex_preview.rb
@@ -1,0 +1,9 @@
+module Lookbook
+  class PhlexPreview < Preview
+    def render(component, &block)
+      super do
+        component.instance_exec component, &block
+      end
+    end
+  end
+end
--- a/spec/dummy/app/views/phlex/card_example.rb
+++ b/spec/dummy/app/views/phlex/card_example.rb
@@ -1,0 +1,11 @@
+module Views::Phlex
+  class CardExample < Phlex::HTML
+    def template(&content)
+      div class: "card", &content
+    end
+
+    def footer(&content)
+      div class: "card-footer", &content
+    end
+  end
+end
--- a/spec/dummy/test/components/previews/phlex_dsl_example_preview.rb
+++ b/spec/dummy/test/components/previews/phlex_dsl_example_preview.rb
@@ -1,0 +1,18 @@
+class PhlexDslExamplePreview < Lookbook::PhlexPreview
+  def default
+    render Views::Phlex::CardExample.new do |card|
+      h1 { "My card" }
+
+      div class: "card-body" do
+        render Views::Phlex::ListExample.new do |list|
+          list.item { "Hello" }
+          list.item { "World" }
+        end
+      end
+
+      card.footer do
+        span { "Hey ya!" }
+      end
+    end
+  end
+end
--- a/spec/requests/previews_spec.rb
+++ b/spec/requests/previews_spec.rb
@@ -45,6 +45,17 @@ RSpec.describe "previews", type: :request do
         expect(html.has_css?("li", text: "World")).to be true
       end
     end
+
+    context "Phlex DSL" do
+      it "supports Phlex DSL" do
+        get lookbook_preview_path("phlex_dsl_example/default")
+
+        expect(html.has_css?(".card > h1", text: "My card")).to be true
+        expect(html.has_css?(".card > .card-body > ul > li", text: "Hello")).to be true
+        expect(html.has_css?(".card > .card-body > ul > li", text: "World")).to be true
+        expect(html.has_css?(".card > .card-footer > span", text: "Hey ya!")).to be true
+      end
+    end
   end
 
   context "View partials" do


app/views/lookbook/previews/preview.html.erb
```diff
@@ -1,6 +1,6 @@
 <% if @render_args[:component] %>
   <% if defined?(Phlex::SGML) && @render_args[:component].is_a?(Phlex::SGML) %>
-    <%= raw(@render_args[:component].call(view_context: self, &@render_args[:block])) %>
+    <%= render(@render_args[:component]) { @render_args[:component].instance_eval(&@render_args[:block]) } %>
```
joeldrapper, line 3: Just a stylistic thing, I would make this use `instance_exec` since it's working with a block rather than a string. You could also pass the component instance to the instance_exec for cases where you want to pick up the yielded component.

Otherwise, I think this makes a lot of sense.

joeldrapper:
This does make sense if you're using Lookbook to document Phlex components that you intend to render inside other Phlex components, however, that's not the behaviour you would get from rendering a Phlex component in another context such as an ActionView view.

Perhaps it should be a configuration option for this reason. 🤔

app/views/lookbook/previews/preview.html.erb
stephannv: Nice. I didn't know the instance_exec. It makes sense.

stephannv:
> This does make sense if you're using Lookbook to document Phlex components that you intend to render inside other Phlex components, however, that's not the behaviour you would get from rendering a Phlex component in another context such as an ActionView view.
> 
> Perhaps it should be a configuration option for this reason. 🤔

Yeah, I guess I can create another preview class, eg.:
Lookbook side:
```ruby
# lib/lookbook/phlex_preview.rb
module Lookbook
  class PhlexPreview < Preview
    def render(component, &block)
      super do
        component.instance_exec component, &block
      end
    end
  end
end
```
```erb
<!-- app/views/lookbook/previews/preview.html.erb -->
<% if defined?(Phlex::SGML) && @render_args[:component].is_a?(Phlex::SGML) %>
  <%= render @render_args[:component], &@render_args[:block] %>
...
```

User side:
```ruby
class MyComponentPreview < Lookbook::PhlexPreview # instead of Lookbook::Preview
  def default
    render MyComponent.new do |c|
      c.title { "Hello" }
      h2 { "World" }
      render MyOtherComponent.new
    end
  end
end
```

app/views/lookbook/previews/preview.html.erb
```diff
@@ -1,9 +1,5 @@
 <% if @render_args[:component] %>
-  <% if defined?(Phlex::SGML) && @render_args[:component].is_a?(Phlex::SGML) %>
-    <%= raw(@render_args[:component].call(view_context: self, &@render_args[:block])) %>
-  <% else %>
-    <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
-  <% end %>
+  <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
```
stephannv, line 2: I'm not sure why this was `raw + call`, but changing to `render component, args, block` worked on specs and on my project. 
Using `render` Rails calls `component.render_in(self, &block)`. Ref: https://github.com/rails/rails/blob/main/actionview/lib/action_view/helpers/rendering_helper.rb#L42

joeldrapper:
Having a different preview class for Phlex-context-previews feels better to me, but rather than instance-exec’ing the the block against the component being rendered, it would be better to have the preview itself be a Phlex::HTML object. That’s more consistent with reality.

It’s possible for components to override HTML elements.

```ruby
class A < Phlex::HTML
  def template
    h1 { "Hello" }
    yield
  end

  def h1 = super(class: "from_a")
end

class B < Phlex::HTML
  def template
    render A.new do
      h1 { "World!" }
    end
  end
end
```

Rendering `B` should output the following because the first heading came from `A` and the second heading came from `B`.
```html
<h1 class="from_a">Hello</h1><h1>World!</h1>
```

But if `B` was a preview object using `instance_exec` on `A`, it would output this instead
```html
<h1 class="from_a">Hello</h1><h1 class="from_a">World!</h1>
```

We can’t make `Lookbook::PhlexPreview` inherit from `Phlex::HTML` because it needs to inherit from `Lookbook::Preview`, but the next version of Phlex will provide everything from `Phlex::HTML` as a module you can include instead. It’s probably worth waiting on that.

stephannv:
> Having a different preview class for Phlex-context-previews feels better to me, but rather than instance-exec’ing the the block against the component being rendered, it would be better to have the preview itself be a `Phlex::HTML` object. That’s more consistent with reality.
> 
> It’s possible for components to override HTML elements.
> 
> ```ruby
> class A < Phlex::HTML
>   def template
>     h1 { "Hello" }
>     yield
>   end
> 
>   def h1 = super(class: "from_a")
> end
> 
> class B < Phlex::HTML
>   def template
>     render A.new do
>       h1 { "World!" }
>     end
>   end
> end
> ```
> 
> Rendering `B` should output the following because the first heading came from `A` and the second heading came from `B`.
> 
> ```
> <h1 class="from_a">Hello</h1><h1>World!</h1>
> ```
> 
> But if `B` was a preview object using `instance_exec` on `A`, it would output this instead
> 
> ```
> <h1 class="from_a">Hello</h1><h1 class="from_a">World!</h1>
> ```

It's true, I could reproduce this problem now =(.

 
> We can’t make `Lookbook::PhlexPreview` inherit from `Phlex::HTML` because it needs to inherit from `Lookbook::Preview`, but the next version of Phlex will provide everything from `Phlex::HTML` as a module you can include instead. It’s probably worth waiting on that.

Nice, I did this on my similar lib (Blueprint) and I think it's a great approach
Screenshot 2024-02-01 at 11 17 31 AM

app/views/lookbook/previews/preview.html.erb
allmarkedup: @stephannv the discussion in #400 should help explain the rationale for this, ugly as it is!

allmarkedup:
@stephannv thanks for taking the time to put this together. And @joeldrapper thanks for jumping in on the discussion with your expertise on this, much appreciated. I'm sadly not using Phlex on a day-today basis at the moment so forgive me if i'm a bit out of my depth with the details of what is being discussed here.

I'm certainly not opposed to having a Phlex-specific preview class to inherit from (or a module that can be included in preview classes for Phlex components) if that means previews can be written in a way that feels more natural for Phlex users.

FWIW, the sub-component render issue is actually also present for ViewComponent previews - to render a sub-component within a parent's component's block context you need to use the parent component's render method. You have to do something like:

```rb
def example_preview
  render ExampleComponent.new do |parent|
    parent.render ChildComponent.new
  end
end
```

This is how it works in ViewComponent's own preview system which Lookbook previews are based on (and aim to be 100% compatible with). It's definitely not ideal for all the reasons you mention in #584 but will likely have to stay like that (for VC component previews, anyway).

So is the general feeling here that it's worth waiting until `Phlex::HTML` (or the contents thereof) is made available as an module before taking this any further?

joeldrapper:
> So is the general feeling here that it's worth waiting until Phlex::HTML (or the contents thereof) is made available as an module before taking this any further?

I think that's our best option. It shouldn't be too long now.

cirdes:
@stephannv helped me out, and I was able to make it work by using an initializer:

`config/initializers/lookbook.rb`
```ruby
module Lookbook::PreviewOverrides
  # see https://github.com/ViewComponent/lookbook/issues/584
  def render(component = nil, **args, &block)
    if block
      super { component.instance_exec component, &block }
    else
      super
    end
  end
end

Rails.application.configure { Lookbook::Preview.prepend Lookbook::PreviewOverrides }
```

But I would love to see better Phlex support!

joeldrapper:
> It shouldn't be too long now.

I said this almost a year ago. 😂

We are not extracting the HTML / SVG behaviours into includable modules anymore. But I think we have a solution that will work well for this.

`Phlex.html` and `Phlex.svg` now take a block and execute it in a new Phlex context using `instance_exec` as suggested above. This Phlex context also copies each of the instance variables from the block’s receiver, which makes it possible to reference state from outside the block while still dispatching methods to the Phlex DSL.

Local variables are available too, because they are captured with the closure by Ruby. And if you need to dispatch methods to the outer context, the receiver is yielded so that’s possible too.

I can take a stab at a new PR based on this one to use these new methods. They were originally proposed here https://github.com/phlex-ruby/phlex/pull/855 but we ended up shipping a slightly different implementation here https://github.com/phlex-ruby/phlex/pull/860.

PR #602, opened 2024-03-14
fix dynamic param value contains double-quotes make alpine x-data has…

xiaohui-zhangxh:
we can simply pass string to x-data with `.to_json`. The old code broken when I type double-quotes in dynamic param text field.

### DIFF
--- a/app/components/lookbook/base_component.rb
+++ b/app/components/lookbook/base_component.rb
@@ -33,11 +33,7 @@ module Lookbook
     end
 
     def alpine_encode(data)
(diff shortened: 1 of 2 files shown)


allmarkedup:
Hey @xiaohui-zhangxh I don't know why it has taken me so long to look at this! Sorry about that.

I believe the issue with quotes in params text has been resolved since this was opened so I'm going too close this as I don't think it is needed anymore. If that is incorrect then let me know and I'll take another look at it.

Many thanks!

PR #617, opened 2024-05-21
Improve responsiveness of UI when navigating

allmarkedup:
Implements a smarter rendering strategy to reduce the amount of HTML sent over the wire when requesting DOM updates.

DOM update responses triggered by navigation requests no longer include re-rendered navigation trees (which can be sizable and slow to render in larger projects) which also means the client-side DOM morphing should be snappier too.

Layouts are also stripped from all Ajax-triggered requests for updated page HTML to further reduce the payload.


camertron (commented):
Hey @allmarkedup, thanks for digging into this! Unfortunately I'm not very well-versed on Lookbook's internals; I usually just fumble around in there until I find the thing that sort of resembles the thing I want to change, then pull out a rusty screwdriver and sort of bang on it until it half works 😅 I understand Alpine even less.

I can offer thanks, but I'm not well equipped to leave a review 😄

allmarkedup:
>  I usually just fumble around in there until I find the thing that sort of resembles the thing I want to change, then pull out a rusty screwdriver and sort of bang on it until it half works

Sounds remarkably like my development strategy 😂 And I certainly don't blame you, it's a bit of a mess in there... I'm working on a big cleanup to hopefully make the codebase a little more sane/approachable to others so maybe one day it will be less terrifying to make changes :)

No worries about the review, I was really just tagging you so you were aware of it. I'm going to merge it now and get a new release out with this and the icon change.

If/when you update Primer to the new version I'd be interested to hear if you think it's made any difference.

camertron:
> it's a bit of a mess in there

I think Lookbook is actually remarkably well-organized and well-written, I'm just personally not very familiar with the internals, or Alpine 😄

Issue #625, opened 2024-06-14
SVG not render in latest Chrome on Mac


michael-yin:
### Describe the bug

In the latest Chrome 126.0.6478.56 on Mac, the svg do not render correctly

### Screenshots

If I visit https://demo.lookbook.build/lookbook

![svg-issue](https://github.com/lookbook-hq/lookbook/assets/4366781/4a0623f5-f2af-4eb2-b45e-7b912b7ad3a6)

### How to fix

lucide-sprite.svg need to be updated

According to https://oliverjam.es/articles/svg-sprites, I changed to `<svg xmlns="http://www.w3.org/2000/svg">` and the issue has been fixed.

Spone:
Hi @michael-yin thanks for reporting this! Would you like to open a PR to apply your fix?

michael-yin:
Hi, @Spone 

I have created a PR to fix the issue, please check it.

Thx.

allmarkedup:
@michael-yin thank youI PR is merged now and I'll get a new release out later today :)

PR #626, opened 2024-06-19
fix svg sprite

michael-yin:
#625


allmarkedup:
This looks good so I'm going to merge it in now - thanks again.

It looks like the underlying issue has been [fixed in Chrome now](https://issues.chromium.org/issues/342579971) but still good to make sure it works for everyone and isn't a problem in the future.

PR #629, opened 2024-06-24
Use dig in layouts guide ERB

neanias:
If one uses the suggested layout for the ViewComponent default layout, navigating to `localhost:3000/rails/view_components` will throw an error due to the deeply nested nature of the Lookbook params hash. Using `dig` allows for safe access into the hashes without causing problems if any of the keys don't exist.

Hash#dig was introduced in Ruby 2.3, so will be available in all supported versions of Lookbook.

There's a few other instances of `params[:some][:deep][:key]` in the docs, but I thought this one may be a better target since it could cross the Lookbook boundary to ViewComponent.

### DIFF
--- a/docs/src/_extend/other/data.md
+++ b/docs/src/_extend/other/data.md
@@ -56,7 +56,7 @@ title: Global Data Store
   <!-- app/views/layouts/preview.html.erb -->
   ...
   <div><%%= yield %></div>
-  <p>Built by <%%= params[:lookbook][:data][:company_name] %></p>
+  <p>Built by <%%= params.dig(:lookbook, :data, :company_name) %></p>
   ...
   ```
 
--- a/docs/src/_guide/previews/display.md
+++ b/docs/src/_guide/previews/display.md
@@ -29,17 +29,17 @@ title: Display Options
     end
     ```
 
-    Display param values can then be accessed via the `params` hash in your preview layout using `params[:lookbook][:display][<key>]`:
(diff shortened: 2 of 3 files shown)


Spone:
Good catch @neanias! Thanks for the PR.

About the docs, it makes sense to improve the syntax there as well, what do you think @allmarkedup?

neanias:
I'm happy to change the others as part of this PR if that helps the sweep?

allmarkedup:
@neanias @Spone sorry about the slow reply! This is a good catch, thank you.

@neanias if you would be willing to update the other places that this syntax is used in the docs that would be fantastic, thank you. If you add it to this PR then I'll merge once ready. Appreciate your time on this!

neanias:
I think that's the main ones. I found [`page.data[:brand_colors][:red]`](https://github.com/lookbook-hq/lookbook/blob/cdcbbaec84bdf9da0577e2a2ee15761d6504e98c/docs/src/_guide/pages/variables.md?plain=1#L45) but I think that one's ok?

Spone (approved):

allmarkedup:
> I think that's the main ones. I found [page.data[:brand_colors][:red]](https://github.com/lookbook-hq/lookbook/blob/cdcbbaec84bdf9da0577e2a2ee15761d6504e98c/docs/src/_guide/pages/variables.md?plain=1#L45) but I think that one's ok?

@neanias yep that should be fine I think. Many thanks for your time with this, will merge now 👍

PR #671, opened 2025-02-24
Fix 667: "Cannot transliterate strings with ASCII encoding" during rails start

everton:
In some environments, if the config files are not explicitly set to use UTF-8, the strings passed to `page_paths` goes encoded as ASCII and the engine pass this down to Ruby's Dir[] method which returns filenames in the same encoding:

```ruby
  Dir['/*'.force_encoding(Encoding::ASCII_8BIT)].map { [_1, _1.encoding] }
  => [
       ["/bin",  #<Encoding:ASCII-8BIT>],
       ["/boot", #<Encoding:ASCII-8BIT>],
       ["/dev",  #<Encoding:ASCII-8BIT>],
       ["/etc",  #<Encoding:ASCII-8BIT>],
       ...
```

```ruby
  Dir['/*'.force_encoding(Encoding::UTF_8)].map { [_1, _1.encoding] }
  => [
       ["/bin",  #<Encoding:UTF-8>],
       ["/boot", #<Encoding:UTF-8>],
       ["/dev",  #<Encoding:UTF-8>],
       ["/etc",  #<Encoding:UTF-8>],
       ...
```

At some point, we pass down these filenames through Rails' String#parameterize, which tries to transliterate it assuming it is a UTF string.

Obs.: instead of testing the utils method isolated, I chose to simulate the real regression by adding the config in the combustion.rb file, and I found another strange behavior there due to the fact that when that file is loaded the Lookbook::Engine already loaded and memoized the page_paths config with the standard "test/components/docs" path on it, it doesn't honor what is in that file for the page_paths config, so I ended up forcing the string that was already in the Engine.

### DIFF
--- a/lib/lookbook/support/utils/utils.rb
+++ b/lib/lookbook/support/utils/utils.rb
@@ -12,7 +12,7 @@ module Lookbook
       end
 
       def name(str)
-        str.to_s.parameterize.tr("-", "_")
+        str.to_s.force_encoding("UTF-8").parameterize.tr("-", "_")
       end
 
       def symbolize_name(str)
--- a/spec/support/combustion.rb
+++ b/spec/support/combustion.rb
@@ -11,8 +11,23 @@ Combustion.initialize! :action_controller, :action_view do
   config.lookbook.project_name = "Lookbook Test App"
   config.lookbook.listen = false
   config.lookbook.using_view_component = true
-
   config.action_controller.default_url_options = {host: "localhost"}
 
   Lookbook.add_tag(:customtag)
+
+  # Regression case with exception:
+  #
+  #   "Cannot transliterate strings with ASCII-8BIT encoding (ArgumentError)"
+  #
+  # In order to simulate this bug we need to force ASCII for filenames (when we
+  # pass an ASCII string to Ruby's Dir[] method it returns filenames as ASCII,
+  # and in some environments, if the code is not explicitly forcing UTF-8 this
+  # will be the case for regular strings ion config files.
+  #
+  # Obs.: At this point here the Lookbook::Engine is already loaded and it
+  #   momoizes the page_paths, so it DOES NOT WORK by adding the
+  #   config.lookbook.page_paths here
+  Lookbook::Engine.page_paths.each do
+    _1.force_encoding(Encoding::ASCII_8BIT)
+  end
 end
--- a/spec/support/combustion.rb
+++ b/spec/support/combustion.rb
@@ -26,7 +26,7 @@ Combustion.initialize! :action_controller, :action_view do
   # will be the case for regular strings ion config files.
   #
   # Obs.: At this point here the Lookbook::Engine is already loaded and it
-  #   momoizes the page_paths, so it DOES NOT WORK by adding the
+  #   memoizes the page_paths, so it DOES NOT WORK by adding the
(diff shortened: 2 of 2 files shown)


spec/support/combustion.rb
```diff
@@ -11,8 +11,23 @@
   config.lookbook.project_name = "Lookbook Test App"
   config.lookbook.listen = false
   config.lookbook.using_view_component = true
-
```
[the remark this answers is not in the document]
everton: It was removed by mistake; I Just added it back! Thank you, sir!

spec/support/combustion.rb
```diff
@@ -15,4 +15,20 @@
...
+  #
+  #   "Cannot transliterate strings with ASCII-8BIT encoding (ArgumentError)"
+  #
+  # In order to simulate this bug we need to force ASCII for filenames (when we
+  # pass an ASCII string to Ruby's Dir[] method it returns filenames as ASCII,
+  # and in some environments, if the code is not explicitly forcing UTF-8 this
+  # will be the case for regular strings ion config files.
+  #
+  # Obs.: At this point here the Lookbook::Engine is already loaded and it
+  #   momoizes the page_paths, so it DOES NOT WORK by adding the
```
Spone, line 29:
```suggestion
  #   memoizes the page_paths, so it DOES NOT WORK by adding the
```

allmarkedup:
@everton Thank you for this! Looks great. I'll merge it in now and it'll go out in the next release. Much appreciated 🙂

lib/lookbook/support/utils/utils.rb
```diff
@@ -12,7 +12,7 @@ def temp_id(prefix: nil)
       end
 
       def name(str)
-        str.to_s.parameterize.tr("-", "_")
+        str.to_s.force_encoding("UTF-8").parameterize.tr("-", "_")
```
chaadow, line 15: sorry but this is not the way to go as it emits a lot of ruby warnings on ruby 3.4
image

```
/3.4.0/gems/lookbook-2.3.7/lib/lookbook/support/utils/utils.rb:15: warning: string returned by :preview.to_s will be frozen in the future
```

chaadow: hopefully this will be reverted or fixed, in the meantime i'll monkey patch this method in a rails initializer @allmarkedup

Spone: Hi @chaadow thanks for reporting! Can you open a PR to fix the warning?

everton: Thanks for reporting it, I was not aware that Symbol#to_s was about to start returning frozen strings.

I believe we can solve it by just calling dup in the string returned after to_s, but I will try to replicate the issue in one test and provide a new PR.

Regarding your suggestion about reverting the original PR, I strongly suggest to not do this, since the original bug was blocking people (including me) to start their applications.


(if you are curious, this is an old issue regarding to_s; You can find a little bit of the history around it here: https://engineering.appfolio.com/appfolio-engineering/2019/10/24/symboltos-returns-frozen-string-in-ruby-27)

everton: Just for the reference, the change on Ruby was added here: https://bugs.ruby-lang.org/issues/20350

PR #672, opened 2025-02-24
support the correct params for phlex v2 call method

fusion2004:
I'm not sure if this is exactly the best approach, and we should probably have appraisal run against phlex v1 & v2 if we want to support both, but this works.

Ideally, I wanted to check against `Phlex::VERSION`, but it seems that isn't autoloaded, so we'd have to put a
```ruby
require "phlex/version" if defined?(Phlex)
```
somewhere.


fusion2004:
Ideally, I wanted to check against `Phlex::VERSION`, but it seems that isn't autoloaded, so we'd have to put a
```ruby
require "phlex/version" if defined?(Phlex)
```
somewhere.

PR #687, opened 2025-03-14
Render ungrouped scenarios in a single pass to support content_for

### LINKED ISSUE #670: Render ungrouped scenarios in a single pass to support content_for
### Describe the bug

ERB templates that utilize `content_for` in the view and `yield` in the layout are not rendered correctly due to the two-step rendering process.

### To Reproduce

1. Specify a `preview_layout` that utilizes named yields, e.g.

```ruby
<% if content_for?(:title) %>
      <%= yield(:title) %>
<% end %>
```

2. Render a template that specifies content:

```ruby
<% content_for :title, "Hi I am a title" %>
```

3. Visit the Lookbook preview
4. The content will be missing.

### Expected behavior

Scenarios rendered on their own (i.e. not as a scenario group) should be able to specify named content and have it reflected through the layout.

### Version numbers 

Please complete the following information:

* Lookbook: 2.3.4
* Rails: 8.0.1
* Ruby: 3.4.2

### Additional context

Relates to https://github.com/lookbook-hq/lookbook/discussions/646.

I've implemented a custom workaround for now on a fork, but would like to confirm the change before looking into testing/supporting embedded previews and opening a PR:

https://github.com/lookbook-hq/lookbook/compare/main...kyeah:lookbook:kev/render-content-for

In addition to this change, I can imagine a few other features that might be desirable, but those are outside the scope of this issue:
- Supporting a different layout option for grouped scenarios
- Supporting a global or per-scenario layout option so that each scenario within a group can be rendered with a specific layout

kyeah:
## Issue

Fixes #670.

## Context

ERB templates that utilize content_for in the view and yield in the layout are not rendered correctly due to the two-step rendering process.

For ungrouped scenarios, the two-step render is unnecessary and we can render the view within the correct layout on the first pass. This allows us to specify named content in the layout using `content_for` and have it reflected correctly.

## Testing

- Added a unit test
- Verified locally that content_for works correctly with this branch
- Verified that updates to the scenario/preview are still hot-reloaded appropriately when changes are made


kyeah:
Hi @Spone @allmarkedup — wanted to check in on this and see if this is a change we could review 👋 We've been wanting to use Lookbook on our project, but this issue prevents us from doing so with our current architecture.

Thank you!

Spone:
Hi @kyeah thanks for contributing! I'll let @allmarkedup do the code review since I'm not very familiar with that part of the codebase yet. Thanks for your patience ❤️

allmarkedup:
Hey @kyeah, thanks for this and sorry for the delay, had a busy week but I'll try and look at this on the weekend 👍

allmarkedup (changes requested):
Hey @kyeah, thanks again for this and sorry It's taken me so long to look at it properly.

Whilst I can understand the need to be able to use `content_for` in partials and have that work correctly in Lookbook, I do have some reservations about the changes in this PR.

I'm still trying to think through all the implications so will share my thoughts below to help get things clear in my head and to try to explain my concerns so that I can figure out what the best course of action is here. 

---

#### 1) Breaking 'component isolation'

The render process was actually intentionally implemented as a two-step process to enforce the concept of component isolation, i.e. that components should be self-contained and not have any effects outside of their own scope.

However, I realise that adding support for rendering ActionView partials/templates in v2 has muddied this water a little because it is now possible to render things that are not strictly _components_ and for which there are good reasons to want to use standard Rails features like `content_for`. So whilst this change does break strict component isolation I'm not too worried about this I don't think.

#### 2) Divergence of supported features for single vs grouped previews

This one gives me a little more pause. Up until now both standalone scenarios and grouped scenarios both supported the same set of features, which means that scenarios can be added to a removed from a group without any side-effects. With this change however, `content_for` would only work within ungrouped scenarios, which could break layouts that rely on named yields if a previously ungrouped scenario was moved into a group.

This doesn't sit quite right with me if I'm honest. If these changes are made then this is something that would definitely need to be covered in the docs. I guess ideally things like  `content_for` would work for grouped scenarios too although I'm not sure how useful that would actually be (and could be quite confusing if it were even possible to implement). 

#### 3)  It's a (potentially) breaking change

If anyone has overridden the `lookbook/previews/group.html.erb` template ([as described in the docs](https://lookbook.build/guide/previews/groups#customising)) and amended the 'single scenario' branch of the conditional then that customisation will now no longer be respected. I think that this is quite an unlikely thing to do (customisation of this file was always really intended to be focused on the group rendering) but it's still something that we need to bear in mind and account for.

---

My gut feeling on this is that it's not a change I'd want to push onto all users, because of the potential for breaking existing Lookbook installations and because the different handling/feature support for grouped/ungrouped scenarios feels just a little off to me.

However... I'm not opposed to putting this behind an opt-in configuration option, as I do see that for certain use-cases being able to use standard Rails features like `content_for` could be useful. Something like `config.lookbook.preview_single_pass_rendering = true` (there might well be a better name for it though!).

@kyeah how does that sound to you? Does the above all make sense?

@Spone do you have any thoughts on potentially putting this behind a config option or any comments on my ramblings above? 

I've also just left one comment on a change that would need to be added in to make sure preview embeds do not break when using the 'single pass' rendering approach.

kyeah:
@allmarkedup Thank you! I appreciate the deep dive and breakdown — I agree with all of those points and some of the reservations. Point (2) crossed my mind, but I had been thinking about it as more of a "partial enhancement" and hadn't considered the confusion that folks may have if they moved an ungrouped scenario into a group.

I also hadn't realized that the group template was customizable, so that makes sense as well regarding point (3).

Based on your assessment, I agree that moving it behind a configuration option makes the most sense, so folks are explicitly opting into the feature and also acknowledging the caveats that we'd document alongside it. Will wait for @Spone's input, but if that approach sounds reasonable to everyone I can take a pass at making those updates.

Spone:
I agree with the config option approach. It would be great if the config option docs explain the potential risks/drawbacks of enabling the option.

allmarkedup:
@kyeah this is looking great now, thank you. 

> Here's what the docs look like now, but open to some copy-editing; I found it a little difficult to concisely explain the customized group.html.erb preview template drawbacks since it only affects the "single-scenario" portion of it.

Yes it's not really an easy thing to explain without more context! I'll have a think but I'm happy to ship it like this for now and then the wording can always be improved later if needs be.

I've got a busy day today but I will do a final check over tomorrow as I'll have a bit more time. And then I'll get it merged and out in the next release if it's all good :)

Spone:
> Yes it's not really an easy thing to explain without more context!

Maybe the docs can include a link to [this comment](https://github.com/lookbook-hq/lookbook/pull/687#pullrequestreview-2762178899) for additional context?

kyeah:
I attempted to add a link in the config option docs, but I wasn't able to follow the <a> tag (everything looks correct, but the click event seems to be swallowed by something 🤔)

allmarkedup (approved):
Okay I've had another check over this and I'm happy for it to be merged now, it's looking good 🎉

> I attempted to add a link in the config option docs, but I wasn't able to follow the tag (everything looks correct, but the click event seems to be swallowed by something 🤔)

Yeah the docs can be a bit of a pain... let's just get this merged and then I can look at making some tweaks to the wording and add in a link as a follow up, it doesn't need to delay this any more I don't think.

I will try to get a new release out this evening with these changes in - I just want to try and get a few fixes for some other issues ready first and then I'll get a new version out.

Thanks again for all your work (and your patience) on this @kyeah!

allmarkedup:
I've just released v2.3.9 which includes this change :)

PR #689, opened 2025-03-28
Fix height style for notes panel in Lookbook inspector

Gianluska:
Just a silly styling fix that was bothering me xD

Before:
The container was set to a fixed size, so when the notes had more than a few lines, they were overflowing, messing up the spacing and losing the background
image

After:
The container height is now set to fit the content, so spacing and background stay intact
image


Spone:
Looking good, thanks!

PR #708, opened 2025-06-03
Support ViewComponent v4

joelhawksley:
In https://github.com/ViewComponent/view_component/releases/tag/4.0.0.alpha6, several ViewComponent APIs related to previews were changed. I've added conditional blocks to support the old and new APIs.

In https://github.com/ViewComponent/view_component/releases/tag/4.0.0.alpha3, ViewComponent no longer subclasses ActionView::Base, meaning we needed to explicitly `include ERB::Util` for `json_escape`.

I updated `Appraisals` to test against ViewComponent versions before and after the breaking changes to ensure backward compatibility.

### DIFF
--- a/docs/src/_guide/previews/index.md
+++ b/docs/src/_guide/previews/index.md
@@ -73,7 +73,7 @@ title: Previews Overview
   <% end %>
 
   <% s.with_block_subheading "Passing parameters", id: "parameters" %>
-  
+
   <% s.with_block_prose do %>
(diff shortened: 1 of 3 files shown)


allmarkedup:
Hey @joelhawksley, thanks for this and sorry again for the slow reply. I've taken a look and it all seems good so I'll merge it in now and get a new release out 👍

allmarkedup (approved):

Issue #699, opened 2025-04-03
`lazy_load_previews_and_pages` doesn't work well with Rails 8


joshuay03:
### Describe the bug

I added this config in https://github.com/lookbook-hq/lookbook/pull/641 having tested it against Rails 7.2. However, since we've upgraded to Rails 8 the loading of pages and previews is quite flaky, with a server restart fixing it most of the time.

I suspect this is due to [Rails 8's deferred route loading](https://github.com/rails/rails/pull/52353), but I haven't had a chance to dig into it yet.

### Expected behavior

With the config enabled, pages and previews should load the same as with it disabled.

### Version numbers 

Please complete the following information:

* Lookbook: 2.3.4
* ViewComponent: 3.21.0
* Rails: 8.0.2
* Ruby: 3.3.7

### Additional context

I am happy to dig into this and send a fix, might not be anytime soon though 😅

allmarkedup:
Hey @joshuay03 - I've spent a little time today looking into this but I'm not seeing the same flakiness on my test setup (which may just be because it's too simple!).

Could you elaborate a little bit on what exactly you mean by "the loading of pages and previews is quite flaky"?

Do you mean that on the initial load not all of the pages/previews are shown correctly? Or that any changes made are not updating correctly? Or something else?

Any other info you could give me to help me figure out what is going on here would be great. The deferred route loading changes do look like they could be a culprit for causing flakiness in this area but it's hard to say without having a better understanding of the issue.

Thanks for your patience with this!

joshuay03:
> Could you elaborate a little bit on what exactly you mean by "the loading of pages and previews is quite flaky"?
> 
> Do you mean that on the initial load not all of the pages/previews are shown correctly? Or that any changes made are not updating correctly? Or something else?

Previews simply don't show up on load. A server restart seems to solve the problem. Unfortunately I wasn't able to replicate this myself, but a few of my coworkers ran into it a few times. I'm away atm but I'll try to replicate again once I'm back.

joshuay03:
I've re-enabled this config in our application so I'm going to close this off for now and re-open it if we run into any issues again, hopefully with some reproduction steps.

joshuay03:
Alright I think I figured out the issue. I'll send a PR.

joshuay03:
> Alright I think I figured out the issue. I'll send a PR.

https://github.com/lookbook-hq/lookbook/pull/718

PR #718, opened 2025-07-20
Fix preview parsing race condition with `lazy_load_previews_and_pages`

joshuay03:
Closes https://github.com/lookbook-hq/lookbook/issues/699

**Context:** When `lazy_load_previews_and_pages` is enabled, [the initial loading of previews is delayed until they are first requested](https://github.com/lookbook-hq/lookbook/blob/e27bbe852e325f0046c04ec4a04e70b3c74be9ee/lib/lookbook/engine.rb#L239-L253).

**Problem:** The existing preview parsing logic uses an instance variable as a boolean flag to prevent concurrent parsing operations. However, this approach has two issues:
1. It's not truly atomic, creating potential race conditions
2. When a second thread attempts to parse while another is already parsing, it simply skips parsing entirely rather than waiting

This race condition is primarily exposed when `lazy_load_previews_and_pages` is enabled. If a "parse changed files" operation is already running when an initial parse request comes in, the initial parse gets skipped entirely. Since no files have actually changed, nothing gets parsed, resulting in an empty preview collection.

Without this config, you'd need to be reloading while preview files are changing to encounter the same issue.

**Solution:** Replace the boolean flag with a mutex to ensure:
- Only one parsing operation executes at a time (thread-safe)
- Queued parse requests are executed rather than skipped, keeping the collection up to date

### DIFF
--- a/lib/lookbook/preview_parser.rb
+++ b/lib/lookbook/preview_parser.rb
@@ -6,15 +6,14 @@ module Lookbook
       @paths = paths
       @after_parse_callbacks = []
       @after_parse_once_callbacks = []
-      @parsing = false
+      @parsing = Mutex.new
 
       define_tags(tags)
       YARD::Parser::SourceParser.after_parse_list { run_callbacks }
     end
 
     def parse(files = nil, &block)
-      unless @parsing
-        @parsing = true
+      @parsing.synchronize do
         @after_parse_once_callbacks << block if block
         files_list = files ? files.select { |file| file.to_s.end_with?(".rb") } : paths
 
@@ -43,7 +42,6 @@ module Lookbook
     def run_callbacks
       callbacks.each { |cb| cb.call(YARD::Registry) }
       @after_parse_once_callbacks = []
-      @parsing = false
     end
 
     def define_tags(tags = nil)


lib/lookbook/preview_parser.rb
```diff
@@ -43,7 +43,6 @@ def callbacks
     def run_callbacks
```
joshuay03, line 43: This will be called with the mutex held because yard runs the `after_parse_list` in the `YARD.parse` call:
- https://github.com/lsegal/yard/blob/5b93b3a4f15a52f3165edef112e718149d294c38/lib/yard/parser/source_parser.rb#L373
- https://github.com/lsegal/yard/blob/5b93b3a4f15a52f3165edef112e718149d294c38/lib/yard/parser/source_parser.rb#L114

lib/lookbook/preview_parser.rb
```diff
@@ -6,14 +6,14 @@ def initialize(paths, tags = nil)
...
+      @parsing = Mutex.new
 
       define_tags(tags)
       YARD::Parser::SourceParser.after_parse_list { run_callbacks }
     end
 
     def parse(files = nil, &block)
-      unless @parsing
+      @parsing.synchronize do
         @parsing = true
```
allmarkedup, line 17: @joshuay03 I think this line needs to be removed doesn't it? Otherwise it overwrites the mutex?

joshuay03: Oops yes. Nice catch!

joshuay03: Fixed 👍🏽

allmarkedup (approved):
Looks good now, many thanks! 🙏

PR #722, opened 2025-07-26
Fix headings IDs in markdown pages

### LINKED ISSUE #721: Markdown headers should produce IDs for anchors
### Is your feature request related to a problem? Please describe

Currently there is no way to link to a specific header within a page. For example, within my component documentation I would like to be able to link to a specific header within my "Best Practices" page.

### Describe the solution you'd like

I'd like Markdown headers in pages to produce anchor tags.

### Additional context

It looks like you're already using the `with_toc_data` option from RedCarpet, so I'm not sure why IDs are not being outputted from headers in the markdown already.

Spone:
I needed to separate Redcarpet's markdown "[extensions](https://github.com/vmg/redcarpet?tab=readme-ov-file#and-its-like-really-simple-to-use)" (such as `tables` and `fenced_code_blocks`) from "[render options](https://github.com/vmg/redcarpet?tab=readme-ov-file#darling-i-packed-you-a-couple-renderers-for-lunch)" (such as `with_toc_data`).

This is a breaking change since Lookbook's `markdown_options` config is renamed `markdown_extensions`. Or should we keep a single `markdown_options`, but automatically filter out `extensions`?

What do you think @allmarkedup?

Close #721

### DIFF
--- a/config/app.yml
+++ b/config/app.yml
@@ -35,16 +35,13 @@ shared:
   page_extensions: [html.*, md.*]
   page_options: {}
 
-  markdown_extensions:
+  markdown_options:
     tables: true
     fenced_code_blocks: true
     disable_indented_code_blocks: true
     strikethrough: true
     highlight: true
     lax_spacing: true
-
-  markdown_options:
-    with_toc_data: true
 
   highlighter_options:
     theme: github
--- a/docs/src/_data/config_options.yml
+++ b/docs/src/_data/config_options.yml
@@ -207,28 +207,28 @@ theming:
       Customise or disable the favicon served by Lookbook.
 
 markdown:
-  - name: markdown_extensions.tables
+  - name: markdown_options.tables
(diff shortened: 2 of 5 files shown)


allmarkedup:
Thanks for this @Spone :)

> This is a breaking change since Lookbook's markdown_options config is renamed markdown_extensions. Or should we keep a single markdown_options, but automatically filter out extensions?

I'd rather avoid a breaking change here if we can so I'd be more in favour of the filtering option. We can then always change the config names as part of the next major version bump. Does that sound okay to you?

Spone:
Yes, definitely. I actually noticed that `with_toc_data` is not mentioned in Lookbook docs, so we can consider this a private API and just remove it from the `markdown_options` (but force the option to `true` internally).
I'll update the PR.

Spone:
Here you go @allmarkedup

PR #795, opened 2026-04-26
Remove unreferenced assets

rwojsznis:
Those asses are part of gem which bloats its size a bit.

Ideally `build` command should do initial cleanup of public folder - but that's out of scope of this PR.

---

Hey there 👋  as per comment - seems like some lost assets are part of the gem - noticed this as I was looking way to shave some MBs of my CI gemfile cache ;)

after the change:

```
ls -lgo public/lookbook-assets/
total 1248
drwxr-xr-x@ 5      160 Apr 26 21:13 css/
drwxr-xr-x@ 3       96 Apr 26 21:13 img/
-rw-r--r--@ 1   245036 Apr 26 21:12 Inter-italic.var.69eb0fe1.woff2
-rw-r--r--@ 1   227180 Apr 26 21:12 Inter-roman.var.b695afbe.woff2
drwxr-xr-x@ 8      256 Apr 26 21:13 js/
-rw-r--r--@ 1    72064 Apr 26 21:12 SourceCodeVariable-Italic.ttf.09b4354a.woff2
-rw-r--r--@ 1    86028 Apr 26 21:12 SourceCodeVariable-Roman.ttf.118e9f22.woff2
```

I'm also unsure about sourcemaps which are generated for both js and css files - maybe those should/could be yanked as well?


rwojsznis:
Hey @Spone (sorry for random ping, seems you were the last who merged few PRs into `main`) - is this repo still maintained? I see few PRs hanging without any feedback for a bit (just askin'! 😄 )

I saw that somebody also tried to ping random collaborators in https://github.com/lookbook-hq/lookbook/pull/799 for a while 😅

Spone:
Hey @rwojsznis thanks for the PR, a smaller gem size is always a good idea!
I double-checked that the fonts are not referenced in the CSS.

PR #799, opened 2026-06-15
chore(deps): allow rouge 5.x

phikai:
## Summary

Widens the Rouge constraint from `< 5.0` to `< 6.0` so applications can pick up [Rouge 5.0](https://github.com/rouge-ruby/rouge/blob/master/CHANGELOG.md), released 2026.

## Compatibility

Rouge 5 dropped Ruby 2.7 and overhauled HTML formatter delegation (rouge-ruby/rouge#2286). Lookbook's only Rouge integration — `Lookbook::CodeHighlighter` — subclasses `Rouge::Formatters::HTML` directly and uses `Lexer.find` / `formatter.format(lexer.lex(...))`, none of which are affected by the delegation overhaul (the Rouge PR explicitly notes "users who have written their own formatters should not be affected").

The existing `Lookbook::CodeHighlighter` spec and the full `bundle exec rspec` suite (587 examples) pass on both Rouge 4.6 and Rouge 5.0 locally with no source changes.

## Notes

- `lookbook.gemspec` still declares `required_ruby_version = ">= 2.7.0"`. Apps on Ruby 2.7 will continue to resolve to Rouge 4.x automatically (Rouge 5 requires Ruby 3.0+), so this change is backward-compatible for the existing supported Ruby range.
- No \`CHANGELOG.md\` in the repo, so nothing to update there.


Spone:
Hi there, sorry for the late reply. What about dropping the upper constraint entirely?

lookbook.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "railties", ">= 5.0"
   spec.add_dependency "view_component", ">= 2.0"
   spec.add_dependency "redcarpet", "~> 3.5"
-  spec.add_dependency "rouge", ">= 3.26", "< 5.0"
+  spec.add_dependency "rouge", ">= 3.26", "< 6.0"
```
Spone:
```suggestion
  spec.add_dependency "rouge", ">= 3.26"
```

Spone:
Thank you for the PR @phikai!

PR #801, opened 2026-06-30
Fix XSS in display options field alpine data interpolation

### LINKED ISSUE #802: DOM-based XSS via _display query parameter on inspect page
### Describe the bug

On the component inspect page, the `_display` query parameter is parsed as JSON and interpolated into the Alpine.js `x-data='...'` attribute without escaping embedded single quotes. A crafted value can break out of the string context and inject arbitrary expressions, which Alpine evaluates via `new AsyncFunction()`.

The value is also persisted in the `lookbook-display-theme` cookie, so the payload keeps firing on subsequent inspect-page navigations.

### To Reproduce

1. Run a Lookbook-hosting app locally.
2. Open any inspect page with:

   ```
   http://localhost:3000/lookbook/inspect/<any-component>?_display=%7B%22theme%22%3A%20%22%27%7D%29%2Calert%281%29%2C%28%7B%27x%27%3A%27%22%7D
   ```

   Decoded `_display`: `{"theme": "'}),alert(1),({'x':'"}`

3. `alert(1)` fires on load.

### Additional context

I created a PR that aims to tackle this issue: https://github.com/lookbook-hq/lookbook/pull/801

liuliu-dev:
Related issue: https://github.com/github/primer/issues/6759

Closes https://github.com/lookbook-hq/lookbook/issues/802

Similar to: https://github.com/lookbook-hq/lookbook/pull/668

Fixes an XSS vulnerability in `Lookbook::DisplayOptions::Field::Component#alpine_data`, where the `name` and `value` of a display option were interpolated directly into an Alpine.js `x-data` attribute without escaping.

Opening a public PR here since the overall risk is `low to none` (the affected surface is the local Lookbook dev UI, with no auth/sensitive data on the reported deployment).

### DIFF
--- a/app/components/lookbook/display_options/field/component.rb
+++ b/app/components/lookbook/display_options/field/component.rb
@@ -17,7 +17,9 @@ module Lookbook
       protected
 
       def alpine_data
-        "{name: '#{name}', value: '#{value}'}"
+        escaped_name = html_escape(json_escape(name.to_s))
+        escaped_value = html_escape(json_escape(value.to_s).gsub("\n", '\n'))
+        "{name: \"#{escaped_name}\", value: \"#{escaped_value}\"}"
       end
 
       def alpine_component


siddharthkp (approved):

lesliecdubs:
Hi @Spone, I see you've approved some recent PRs in this repo - could you help us out on a review for this one? We'd love to get this patched!

app/components/lookbook/display_options/field/component.rb
```diff
@@ -17,7 +17,9 @@ def choices
       protected
 
       def alpine_data
-        "{name: '#{name}', value: '#{value}'}"
+        escaped_name = html_escape(json_escape(name.to_s))
+        escaped_value = html_escape(json_escape(value.to_s).gsub("\n", '\n'))
+        "{name: \"#{escaped_name}\", value: \"#{escaped_value}\"}"
```
Spone: **Residual robustness gap (not a security issue).** `json_escape` and `html_escape` both leave backslashes untouched, and this now uses `"` as the JS-string delimiter. A value ending in an odd number of backslashes (e.g. a lone `\`) renders as `value: "\"}` → an unterminated JS string literal → Alpine `x-data` parse error, silently breaking this field. Not exploitable (every attacker-supplied quote is neutralized to `&quot;`/`&#39;`, so no code execution), but it's a real correctness bug I hit while fuzzing the fix.

**Suggested robust form:** stop hand-interpolating and let a serializer handle quotes/backslashes/U+2028-2029 uniformly, e.g. build a Hash and `html_escape(json_escape({name:, value:}.to_json))`. Note this same fragile pattern was added in #668 (params field), and `BaseComponent#alpine_encode` has the same quote bug — a shared `.to_json`-based helper would fix all three sites at once. Fine to land this targeted fix now and do the consolidation as a follow-up.

Spone (commented):
I reproduced the reported XSS on `main` and confirmed this patch closes it. Rendering the exact pipeline (this method → Rails `tag` builder attribute escaping → single-pass browser entity-decode → Alpine `x-data` JS eval) against actionview 8.1.3.1:

- **single-quote breakout** `x'}, evil:alert(1), z:{a:'` — on `main`, Alpine ends up evaluating `{value: 'x'}, evil:alert(1), …}` and `alert(1)` fires. With this patch the `'` lands as inert `&#39;` text inside the `"`-delimited string. ✅
- double-quote breakout and `</script>`/angle-bracket payloads are neutralized too. ✅

The mechanism is correct but subtle: it only holds because the Rails tag builder applies a *second* html_escape on top of the manual one, so one entity layer survives the browser's single decode pass. Worth a comment in the code so nobody "simplifies" away the manual `html_escape` later. Two notes below.

app/components/lookbook/display_options/field/component.rb
Spone: Fixed in 375de4b7bf3240fa91858452e54ac3eecaafc2f5

Spone:
I pushed a follow-up commit that reworks this fix — flagging the scope change so it's not a surprise.

**Why it grew:** verifying against the actual rendered component (rather than a bare `tag.div`) surfaced that the original approach here was still exploitable. `Lookbook::TagComponent` renders attributes with `escape: false`, so on the `render_component_tag` path the manual `html_escape` was the *only* escaping layer. That's fine for a single quote, but a **double quote** in the value round-trips like this:

- component emits `value: &quot;x&quot;}, …&quot;`
- the browser decodes the attribute once → `value: "x"}, …"`
- the attacker's `"` is now a real quote → it closes the JS string early and the trailing text is evaluated as an Alpine expression.

So the breakout was only *narrowed*, not closed.

**The fix:** route both this field and the preview-params field (the sibling fix from #668, which had the same hand-rolled pattern) through `BaseComponent#alpine_encode`, rewritten to serialise with `to_json`. `to_json` backslash-escapes quotes *inside* string values (`\"`), and that escape survives the browser's single entity-decode regardless of how the attribute is HTML-escaped — so the value can never terminate the JS string. Verified safe in **both** render contexts (`escape: false` here, `escape: true` on the params path), which is what lets a single helper cover every call site.

Key detail captured in a code comment: **don't** wrap `alpine_encode` in `html_escape` — on the `escape: true` path the tag builder escapes again and you get literal `&quot;` inside the JSON, breaking the payload.

Added `spec/components/alpine_encode_spec.rb` — renders both components against single / double / mixed quotes, trailing backslash, `</script>` and line-separator payloads, asserting each value round-trips through entity-decode + `JSON.parse`. Full suite stays green (605 examples), including the five other `alpine_encode` consumers (nav, copy button, dimensions, embed), so nothing regressed.

